PR #21, opened 2022-03-15
Update reporting a vulnerability documentation

michaeljmarshall:
See https://github.com/apache/pulsar/pull/14610 and https://github.com/apache/pulsar/pull/14610#issuecomment-1067510855 for context.

Adding some detail to the reporting a vulnerability documentation.

@Anonymitaet - I noticed that we have a contact page on the new website https://pulsar-next.staged.apache.org/contact/. I propose adding this contact information on both pages (the "how to contribute" page and the "contact" page).

### DIFF
--- a/site2/website-next/src/pages/community.js
+++ b/site2/website-next/src/pages/community.js
@@ -222,6 +222,7 @@ export default function Community(props) {
                   <div className="sm:w-2/3">
                   <h3>Reporting a Vulnerability</h3>
                   <p>To report a vulnerability for Pulsar, contact the <a className="secondary-cta" href="https://www.apache.org/security/projects.html" target="_blank">Apache Security Team</a>.</p>
+                  <p>The process for reporting a vulnerability is outlined <a className="secondary-cta" href="https://www.apache.org/security/" target="_blank">here</a>. When reporting a vulnerability to <a className="secondary-cta" href="mailto:security@apache.org" target="_blank">security@apache.org</a>, you can copy your email to <a className="secondary-cta" href="mailto:private@pulsar.apache.org" target="_blank">private@pulsar.apache.org</a> to send your report to the Apache Pulsar Project Management Committee. This is a private mailing list.</p>
                   </div>
                 </div>
               </div>
--- a/site2/website-next/src/pages/contact.js
+++ b/site2/website-next/src/pages/contact.js
@@ -56,6 +56,23 @@ export default function page(props) {
           </h2>
           <MailTable data={mailingLists}></MailTable>
           <h2>
+            <translate>
+              Reporting a Vulnerability
+            </translate>
+          </h2>
+          <p>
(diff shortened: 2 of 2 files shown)


michaeljmarshall:
I verified that this "looks" correct. Here are screenshots:
Screen Shot 2022-03-15 at 12 20 37 AM
Screen Shot 2022-03-15 at 12 21 01 AM

Anonymitaet:
I think one place (below) is enough since we do not want duplicated content, right? @D-2-Ed @DaveDuggins

image

michaeljmarshall:
@Anonymitaet - I agree that we should probably avoid complete duplication. However, I'll note that the Apache Spark project does point to their "Reporting a Vulnerability" page from the mailing list page.

See https://spark.apache.org/community.html and https://spark.apache.org/security.html

Anonymitaet:
@michaeljmarshall thanks for your explanations, but we do not need to follow the "Spark way" 😄

michaeljmarshall:
@Anonymitaet - I did not mean to suggest that we should follow their way blindly.

In my view, we should make this contact information easy to find, even if it is slightly redundant. I shared the Apache Spark page because it shows the design I have in mind. 

My main point is that the contact page should include information on how to report a vulnerability or it should link a user to that part of our website. Given that our "report a vulnerability" protocol is primarily "send an email to security@apache.org", I think it makes sense on the contact page where we have all relevant Pulsar email addresses.

Beside the obvious benefit of helping security researchers know our protocol, this also ensures that users will notice that we have a well defined security protocol in place.

Note that in some of the examples you shared, the community and the contact pages were joined. I agree with you that it wouldn't make sense to have the information in two places if they shared a single webpage.

What is your perspective? Thanks.

Anonymitaet:
@michaeljmarshall thanks for your explanations! That makes sense.
To avoid duplication and add references, how about changing it as below?
(add a row for security@apache.org in the Mailing Lists table)
image

urfreespace:
@michaeljmarshall @Anonymitaet Any new progress? If there is no agreement within 3 days, I will close this PR, PTAL, thanks.

michaeljmarshall:
@Anonymitaet - that is a good solution. In looking at making the change, I noticed that I cannot update the table without also adding links for "subscribe", "unsubscribe", and "archives". Do you know of a way around that? Thanks!

Anonymitaet:
> @Anonymitaet - that is a good solution. In looking at making the change, I noticed that I cannot update the table without also adding links for "subscribe", "unsubscribe", and "archives". Do you know of a way around that? Thanks!

Sorry, I do not know. If that can not fit into the table, consider add a paragraph after the table?

dave2wave:
Make the names of the list links to the archives. (https://lists.apache.org/list.html?dev@pulsar.apache.org)

On the mailing lists page there is a subscribe button. It won't be translated, but ...

tisonkun:
I'm going to follow up this thread. With point out all this entrypoint to a new security page and add `Security advisories` from the main repo wiki.

tisonkun:
Closing...

Superseded by #345. You're welcome to give it a review.

PR #42, opened 2022-03-31
Added docs and supporting images

DaveDuggins:
Added Core concepts and Core concepts API design topics to /docs. Added related images to static/assets.


Anonymitaet:
Hi @DaveDuggins our convention is: before sending a PR, we run a local preview and then attach the preview screenshots in the PR description, so that others know your changes clearly and quickly. Could u pls do that? and then ping reviewers to review. Thanks.

Example: https://github.com/apache/pulsar/pull/14314

PR #52, opened 2022-04-19
Updated sidebar label tags

DaveDuggins:
See code example.


Anonymitaet:
@DaveDuggins why modify the sidebar and make them look the same?
image

PR #79, opened 2022-05-17
Add filters to the Ecosystem and the Case Studies pages

zrsaber:
Complete two（Add filters to the [Ecosystem](https://pulsar.apache.org/ecosystem) and the [Case Studies](https://pulsar.apache.org/case-studies) pages） from https://github.com/apache/pulsar/issues/15550

image
image
image
image

I removed the original filtering method, because it will cause duplication


urfreespace (approved):
LGTM

PR #123, opened 2022-06-28
Remove old Python Client docs

michaeljmarshall:
## Motivation

The current docs are generated in an additive process where we build from the `main` branch and then copy the result to the `asf-site-next` branch. That means incorrect files are left in the `asf-site-next` branch. The only way to remove branches is manually right now. This PR removes pages that should not be on the website.

Note that this PR partially solves the problem described in https://github.com/apache/pulsar/issues/15984. It does not solve the underlying problem, but it at least makes the python docs discoverable without needing to know where they are.

## Changes

* Remove several versions of Python docs that are `-SNAPSHOT` docs. These directories have corresponding directories that supersede these `-SNAPSHOT` docs. 
* Remove the python docs in the `content/api/python/` directory. The consequence of this change is that this page https://pulsar.apache.org/api/python/ will stop rendering old python docs and will instead render all python doc versions. We should probably fix that experience, but at the very least, this change will fix the underlying problem of having docs that are unreachable.


michaeljmarshall:
Confirming the change had the desired affect. This link https://pulsar.apache.org/api/python/ is now:

Screen Shot 2022-06-27 at 9 50 28 PM

Like I said, we'll probably want to fix the format, but I would rather have something that works for each specific version of the python client than an incorrect doc that isn't for any released version of the python client.

PR #124, opened 2022-06-28
Fix versioned links for clients

michaeljmarshall:
Fixes https://github.com/apache/pulsar/issues/15984.

## Motivation

We have many versioned links in our docs. We replace those links by running `node replace.js` on the script I update in this PR. The versioning is currently broken.

## Change

* Fix the `clientVersionUrl` method in the `website-next/scripts/replace.js` file. Note that this changes the method so that it matches the `website/scripts/replace.js`.

### DIFF
--- a/site2/website-next/scripts/replace.js
+++ b/site2/website-next/scripts/replace.js
@@ -91,24 +91,14 @@ function debDistUrl(version, type) {
 }
 
 function clientVersionUrl(version, type) {
-  return `${siteConfig.url}/api/${type}`;
-  // var versions = version.split(".");
-  // var majorVersion = parseInt(versions[0]);
-  // var minorVersion = parseInt(versions[1]);
-  // if (majorVersion === 2 && minorVersion < 5) {
-  //   return `${siteConfig.url}/api/` + type + `/` + version + "/";
-  // } else if (majorVersion >= 2 && minorVersion >= 5) {
-  //   return (
-  //     `${siteConfig.url}/api/` +
-  //     type +
-  //     `/` +
-  //     majorVersion +
-  //     `.` +
-  //     minorVersion +
-  //     `.0` +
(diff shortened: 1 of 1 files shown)


michaeljmarshall:
Note that I can confirm for the client api that these links are correct. See this link: https://pulsar.apache.org/api/.

michaeljmarshall:
Confirming that the website-next built and works correctly!

michaeljmarshall:
Correction, it works for the 2.10.0 docs. The other docs appear to be the same. I'll continue debugging this.

michaeljmarshall:
After looking closer at the current website, it worked for https://pulsar.apache.org/docs/client-libraries-python/, but not for https://pulsar.apache.org/docs/2.10.0/client-libraries-python/. I double checked that the script works by running `node ./scripts/replace.js` from `site2/website-next`, I can see that the script properly updates all client version links for the explicitly versioned docs. I'm not sure why the update isn't getting saved as a commit to the `asf-site-next` branch.

michaeljmarshall:
@urfreespace - it looks like the website build for versioned docs is not working correctly. I think something might be broken. The change for this PR didn't get applied, and in looking closer, I see that there are other missing changes. For example, if you look at https://pulsar.apache.org/docs/2.10.0/io-use/#source-3, you can see that it says `For more information, see [here](https://pulsar.apache.org/docs/2.10.0/io-cli.md#start).`. However, the current docs in `pulsar-site` at path `site2/website-next/versioned_docs/version-2.10.0/io-use.md` have the same line as `For the latest and complete information, see [Pulsar admin docs](/tools/pulsar-admin/).`. I started to look at this tonight but didn't find the underlying issue. I plan to continue looking at this, too. Thanks for any help you can provide!

urfreespace:
> 

I think @Anonymitaet can provide more context about it

PR #125, opened 2022-06-28
Remove trailing slash from clientVersionUrl

michaeljmarshall:
In #124, I incorrectly included a trailing slash, which led to URLS getting generated with `//` at the end. This PR fixes that mistake.


Anonymitaet:
@urfreespace review?

PR #138, opened 2022-07-06
Only replace correctly formed endpoints

michaeljmarshall:
The current replace.js script leads to malformed output because it converts proper links like `http://pulsar.apache.org/api/client/org/apache/pulsar/client/api/ConsumerConfiguration.html#setPriorityLevel-int-` into `http://pulsar.apache.orghttps://pulsar.apache.org/api/client/2.1.0-incubating/org/apache/pulsar/client/api/ConsumerConfiguration.html#setPriorityLevel-int-`. Instead, we should only replace links when they are properly at the start of a markdown file link, or when they start with a left paren.

You can observe the current behavior and the changed behavior by running `node ./scripts/replace.js` in the `site2/website-next` directory then run `git diff` then grep for `orghttps`. After this change, there are not any malformed links.

```diff
-`priorityLevel` | int | no | Define a [priority](http://pulsar.apache.org/api/client/org/apache/pulsar/client/api/ConsumerConfiguration.html#setPriorityLevel-int-) for the consumer
+`priorityLevel` | int | no | Define a [priority](http://pulsar.apache.orghttps://pulsar.apache.org/api/client/2.1.0-incubating/org/apache/pulsar/client/api/ConsumerConfiguration.html#setPriorityLevel-int-) for the consumer

```


michaeljmarshall:
Note that in the above comment, the replace script tries to create this link: https://pulsar.apache.org/api/client/2.1.0-incubating/org/apache/pulsar/client/api/ConsumerConfiguration.html#setPriorityLevel-int-. That isn't currently a valid link, and the `ConsumerConfiguration` class is deprecated. http://pulsar.apache.org/api/client/org/apache/pulsar/client/api/ConsumerConfiguration.html#setPriorityLevel-int- is currently valid. We need to address these links before migrating them to be replaced with the `replace.js` script.

urfreespace (approved):

PR #148, opened 2022-07-18
Remove locales we don't support yet

tisonkun:
Otherwise when a user change the language but found the result is just English, it's not a good experience.

If we don't support it, we don't mention it. We can always add them back when there is real translated contents.

### DIFF
--- a/site2/website-next/scripts/split-version-build.sh
+++ b/site2/website-next/scripts/split-version-build.sh
@@ -1,7 +1,7 @@
 #! /bin/sh
 node scripts/split-version.js
 
-locals=("en" "zh-CN" "zh-TW" "ja" "ko" "fr")
+locals=("en" "zh-CN")
 latest=$(cat scripts/.latest)
 BUILD_ALL_LANGUAGE="0"
 buildVersion="next"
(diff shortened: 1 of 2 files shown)


Anonymitaet (commented):
@tuhaihe @Sherlock113 do we have any progress on the localization work (zh-tw, ja, fr, and ko)?

tisonkun:
@Anonymitaet thanks for your reaction. Since we don't run any CI tasks to dry run and verify the build, may I ask how to verify all changes looks good locally? I've run `./preview.sh`, but it only generates English version.

cc @urfreespace

Anonymitaet:
@tisonkun double check: do you want to preview localized pages? Seems that they can not be previewed, correct? @urfreespace

tisonkun:
@Anonymitaet I make this patch with intention to avoid display language options which we don't have a real translated content. However, I don't know whether I change it correctly. I'd like to verify it locally, no matter is "localized pages" or other approach.

For example, when I execute `/preview.sh`, I get:

image

.. it seems other options are populated in a following step.

Anonymitaet:
If you want to preview changes locally, then `./preview.sh` is OK and enough

tisonkun:
@Anonymitaet thank you. Then I'm waiting for more review comments.

The original purpose to make this change is "If we don't support it, we don't mention it". Even if we _have a plan_ to offer those translated contents, we can re-add the language switch option when they're present.

tuhaihe:
l10n work are organized via [Apache Pulsar @Crowdin](https://crowdin.com/project/apache-pulsar) project. Some translation progress for different languages for the old website, including ZH, JP, etc. But the new website made all the translation work failure. In the following weeks, I can help contact the contributors to rework on the translation. 

So let's keep them first. No rush to remove.

tisonkun:
@tuhaihe thanks for your information. Then here could be a first step to mention the translation status in the repo or on the site.

Closed as invalid. Though the status should be updated.

PR #169, opened 2022-08-21
Move config docs to /reference and improve docs gen scripts

SignorMercurio:
@urfreespace PTAL carefully.

### DIFF
--- a/site2/tools/build-site.sh
+++ b/site2/tools/build-site.sh
@@ -22,10 +22,12 @@ ROOT_DIR=$(git rev-parse --show-toplevel)
 VERSION=$(${ROOT_DIR}/src/get-project-version.py)
 
 NEXT=$1
-WEBSITE_DIR="website"
+WEBSITE_DIR=${ROOT_DIR}/site2/website
 if [ -n "$NEXT" ]; then
-  WEBSITE_DIR="website-"$NEXT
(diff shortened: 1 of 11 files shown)


site2/tools/build-site.sh
```diff
@@ -22,10 +22,12 @@ ROOT_DIR=$(git rev-parse --show-toplevel)
...
 
 NEXT=$1
-WEBSITE_DIR="website"
+WEBSITE_DIR=${ROOT_DIR}/site2/website
 if [ -n "$NEXT" ]; then
-  WEBSITE_DIR="website-"$NEXT
+  WEBSITE_DIR=${ROOT_DIR}/site2/website-$NEXT
 fi
+TOOLS_DIR=${ROOT_DIR}/site2/tools
+GEN_SITE_DIR=$GEN_SITE_DIR
```
urfreespace, line 30: why `GEN_SITE_DIR=$GEN_SITE_DIR`? Is this redundant? @SignorMercurio

urfreespace (changes requested):
@SignorMercurio good job, thanks for these optimizes, do you have a full build test as you know, it's a big update

site2/tools/build-site.sh
SignorMercurio: This is a mistake. I'll fix it and try to test it locally.

PR #179, opened 2022-08-25
[cleanup][doc] Remove dangling pages

tisonkun:
This refers to https://github.com/apache/pulsar/pull/17265.


tisonkun:
cc @Anonymitaet @urfreespace

urfreespace (approved):

Anonymitaet (approved):

tisonkun:
It seems even if we delete on master, the publish branch still has the content. We don't remove existing content but incrementally add diff to the deploy branch (I think it's `asf-site-next`).

I can push one more follow-up to delete them manually, but we'd better overwrite all content when publishing?

cc @urfreespace @Anonymitaet

Issue #173, opened 2022-08-24
Improve color shades for dark mode


tisonkun:
Currently, in dark mode, the contents on several pages are hard to read.

For example, https://pulsar.apache.org/docs/next/tiered-storage-filesystem#configure-filesystem-offloader-driver.

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/18818196/186356215-8b6e3d0f-0ded-4db4-9be7-d63c518f54bc.png">

cc @Anonymitaet @urfreespace

Anonymitaet:
@tisonkun thanks for raising this up!

Even though the Pulsar website is built in the `pulsar-site` repo and many website-related data are hosted here, I still suggest reporting website-related issues in the `pulsar` repo. 

Reasons:

- High visibility, which means the issue can get more attention since the majority of ppl go to `pulsar`. Only a few ppl know `pulsar-site`.

- From past to now, we nearly track all website-related issues in `pulsar`, it's convenient for us to triage and check them all in one place.

Thoughts?

tisonkun:
@Anonymitaet for this specific issue, I think we need UI experts to help with selecting colors series.

The related code is https://github.com/apache/pulsar-site/blob/e1785ad30a9bd02047ccd31e85e9c89fa1c37c6a/site2/website-next/src/css/custom.css while is hosted primary in this repo instead of synced from `pulsar`.

@urfreespace suggests I report site-related topics here and content-related topics in the main repo. Although, I agree on the "High visibility". I'm OK to just disable the issue tab in this repo and report all issues in `pulsar` - [SkyWalking](https://github.com/apache/skywalking) follows this pattern.

Anonymitaet:
Hi designers, @gancunjie @DaveDuggins, is it a good practice to set all the texts in the dark color background to `white`?

<img width="1077" alt="image" src="https://user-images.githubusercontent.com/50226895/186384398-a6b3c13f-b49b-457a-9895-64740cfe3c48.png">

urfreespace:
fixed with this PR https://github.com/apache/pulsar-site/pull/180

tisonkun:
Thanks! Although the color series seems can be continuously improved, now the content can be read clearly.

tisonkun:
@urfreespace found one case inconsistent: https://pulsar.apache.org/versions/

[...]

PR #180, opened 2022-08-25
update css for table bg color

urfreespace:
https://github.com/apache/pulsar-site/issues/173


site2/website-next/static/swagger/restApiVersions.json
```diff
@@ -81,7 +81,7 @@
             ]
         }
     ],
-    "2.4.0": [
+    "2.4.1": [
```
tisonkun, line 84: Where do these unrelated changes come from?

Issue #175, opened 2022-08-24
Remove all content before sync


tisonkun:
@urfreespace @Anonymitaet Otherwise, we can suffer from some files deleted from upstream but still remain in this repo.

michaeljmarshall:
I think this kind of change will break certain links that I left in place a couple months ago. For example, many of the SNAPSHOT links here https://pulsar.apache.org/api/client/ will go away. I think it is fine to get rid of them if we are ready to do it, but I was slightly worried that users would have saved links to those endpoints, so I didn't remove them.

tisonkun:
@michaeljmarshall then it's an issue about the source of truth. If we retain those content, how can we remove it from the main repo?

The original purpose of this issue is to avoid dangling content under `docs` folder, I may narrow the scope there, though.

tisonkun:
Since docs are versioned, it should not be a big deal. If we'd like to retain the link, we set up a redirection.

michaeljmarshall:
That is a fair point about the source of truth. Maybe we just need to have a landing page? In the case of the API docs, the old content wasn't even based on released versions of Pulsar, so its value is questionable, anyway.

Anonymitaet:
@tisonkun 
Suggest keeping all docs in the `Pulsar` repo, there was a similar discussion showing the benefits of this way.
https://lists.apache.org/thread/mf2xwntfgn84dq78ksqv22jk3drq6xb3

tisonkun:
@Anonymitaet Thanks for your information.

@michaeljmarshall If I get it right, in the mail thread linked your preference is keeping the "current" (unreleased) document in the main repo? Keep released versioned document, especially still retain those like "version-2.1.0-incubating", "version-2.1.1-incubating" really help?

michaeljmarshall:
> If I get it right, in the mail thread linked your preference is keeping the "current" (unreleased) document in the main repo?

I think it is helpful to keep docs that are manually maintained in the apache/pulsar repo. Once a version is no longer supported, I don't think we need those docs in the apache/pulsar repo since the version itself is not getting updated.

> Keep released versioned document, especially still retain those like "version-2.1.0-incubating", "version-2.1.1-incubating" really help?

I support removing these docs from the apache/pulsar repo, and probably from the website, but I am not sure. One of my biggest problems with the old docs is that I still get served them by search engines above more recent docs. We definitely want users to find recent documentation.

tisonkun:
>  that are manually maintained

[...]

PR #183, opened 2022-08-27
[build][site] cleanup previous content before copy new ones

tisonkun:
This refers to #175 and resolves https://github.com/apache/pulsar-site/pull/179#issuecomment-1228636515.

I don't know whether or not it causes some regressions I'm unaware of. Please help with reviews @urfreespace @michaeljmarshall @dave2wave

### DIFF
--- a/.github/workflows/ci-pulsar-website-next-build.yaml
+++ b/.github/workflows/ci-pulsar-website-next-build.yaml
@@ -33,9 +33,9 @@ jobs:
     timeout-minutes: 600
     steps:
       - name: checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
-          fetch-depth: 2
+          fetch-depth: 1
 
       - name: Get changed files
         uses: ./.github/actions/changed-files
@@ -91,7 +91,7 @@ jobs:
 
       - name: Generate swagger json file
         run: mvn -B -ntp -pl pulsar-broker install -DskipTests -Pswagger
-      
+
       - name: Copy swagger json file
         run: |
           echo "Copying swagger json file ..."
--- a/.github/workflows/ci-pulsar-website-next-build.yaml
+++ b/.github/workflows/ci-pulsar-website-next-build.yaml
@@ -35,7 +35,7 @@ jobs:
       - name: checkout
         uses: actions/checkout@v3
         with:
-          fetch-depth: 1
+          fetch-depth: 2
 
       - name: Get changed files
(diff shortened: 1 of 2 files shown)


site2/tools/publish-website.sh
```diff
@@ -52,11 +51,7 @@ PULSAR_SITE_TMP=/tmp/pulsar-site
     mkdir -p $PULSAR_SITE_TMP/content/
   fi
 
-  echo "BUILD_ALL:"$BUILD_ALL
-  if [[ $BUILD_ALL"" == "11" ]]; then
-    echo "clean all the old random js files because a new all version build done, all the js is regenerated with a random name"
-    rm -rf $PULSAR_SITE_TMP/content/assets/js
-  fi
+  rm -rf $PULSAR_SITE_TMP/content/*
```
urfreespace, line 54: you can't remove all content every time, just when the website has a full build, you can do that, in fact, we just build the versions which have files changes, just once full build(all versions and all languages) one day, so these code can't be removed

urfreespace (changes requested):

.github/workflows/ci-pulsar-website-next-build.yaml
```diff
@@ -33,9 +33,9 @@ jobs:
     timeout-minutes: 600
     steps:
       - name: checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
-          fetch-depth: 2
+          fetch-depth: 1
```
urfreespace, line 38: here, we need depth 2, we need to compare the changes of the last 2 commits because we judge which versions of the documents have changed and only build the changed versions

urfreespace (changes requested):

site2/tools/publish-website.sh
tisonkun: It seems a full build takes 40 minutes to complete, while we normally build every six hours. Do we really need this improvement?

The current issue is that if we don't remove content, we can have dangling pages even if it's removed on the master branch.

But yes, we can retain the current logic while removing all content when build all :)

tisonkun: Updated.

.github/workflows/ci-pulsar-website-next-build.yaml
tisonkun: Thanks! Reverted.

tisonkun:
Updated to retain the current logic while removing all content when build all.

urfreespace (approved):

PR #184, opened 2022-08-28
Migrate all docs from /tools to /reference

### LINKED ISSUE #177: Migrate all docs from `/tools` to `/reference`
According to https://apache-pulsar.slack.com/archives/C03MG1H2B3N/p1661220553497289, we can start working on migrate all docs from `/tools` to `/reference`. For instance, https://pulsar.apache.org/tools/pulsar-admin/2.11.0-SNAPSHOT/ will be moved to https://pulsar.apache.org/reference/ and can be viewed in the `Pulsar Admin` section of the side navbar, just like `Pulsar Configuration` right now.

I'm currently working on this.

SignorMercurio:
No longer uses `pulsar*-md.sh` and brodocs because there's no need to generate HTML files. Meanwhile, add separate sidebars for each component and a global navbar to navigate between these components.

Fix #177 

@urfreespace PTAL


urfreespace (approved):

tisonkun:
It seems this patch causes build failed: + /pulsar/site2/tools/pulsar-admin-doc-gen.sh /pulsar/site2/website-next

tisonkun:
@urfreespace we may revert this patch first to unblock site build and digging what cause the failure later.

SignorMercurio:
I see. I'll fix this soon.

PR #207, opened 2022-09-08
fix docs banner text

urfreespace:
## Unreleased version banner
image

## Maintained version banner
image

## Unmaintained version banner
image

### DIFF
--- a/site2/website-next/scripts/code.json
+++ b/site2/website-next/scripts/code.json
@@ -1,0 +1,501 @@
+{
+  "Subscribe": {
+    "message": "Subscribe"
+  },
+  "Unsubscribe": {
+    "message": "Unsubscribe"
+  },
+  "Archives": {
+    "message": "Archives"
+  },
+  "Release Notes": {
+    "message": "Release Notes"
+  },
+  "Documentation": {
+    "message": "Documentation"
+  },
+  "theme.navbar.mobileVersionsDropdown.label": {
+    "message": "Versions",
+    "description": "The label for the navbar versions dropdown on mobile view"
+  },
+  "Contact": {
+    "message": "Contact"
+  },
+  "There are many ways to get help from the Apache Pulsar community. The mailing lists are the primary place where all Pulsar committers are present. Bugs and feature requests can either be discussed on the dev mailing list or by opening an issue on": {
+    "message": "There are many ways to get help from the Apache Pulsar community. The mailing lists are the primary place where all Pulsar committers are present. Bugs and feature requests can either be discussed on the dev mailing list or by opening an issue on"
+  },
+  "Mailing Lists": {
+    "message": "Mailing Lists"
+  },
+  "Stack Overflow": {
+    "message": "Stack Overflow"
+  },
+  "For technical questions, we ask that you post them to": {
+    "message": "For technical questions, we ask that you post them to"
+  },
+  "using the tag “apache-pulsar”.": {
+    "message": "using the tag “apache-pulsar”."
+  },
+  "Slack": {
+    "message": "Slack"
+  },
+  "There is a Pulsar slack channel that is used for informal discussions for devs and users.": {
+    "message": "There is a Pulsar slack channel that is used for informal discussions for devs and users."
+  },
+  "The Slack instance is at": {
+    "message": "The Slack instance is at"
+  },
+  "You can self-register at": {
+    "message": "You can self-register at"
+  },
+  "WeChat": {
+    "message": "WeChat"
+  },
+  "There are several WeChat groups that are used for informal discussions for devs and users.": {
+    "message": "There are several WeChat groups that are used for informal discussions for devs and users."
+  },
+  "To join these WeChat tech groups, you can add Bot with the WeChat ID: StreamNative_BJ": {
+    "message": "To join these WeChat tech groups, you can add Bot with the WeChat ID: StreamNative_BJ"
+  },
+  "Apache Pulsar downloads": {
+    "message": "Apache Pulsar downloads"
+  },
+  "Release notes": {
+    "message": "Release notes"
+  },
+  "Current version (Stable)": {
+    "message": "Current version (Stable)"
+  },
+  "Tiered storage offloaders": {
+    "message": "Tiered storage offloaders"
+  },
+  "Pulsar IO connectors": {
+    "message": "Pulsar IO connectors"
+  },
+  "Release Integrity": {
+    "message": "Release Integrity"
+  },
+  "Getting started": {
+    "message": "Getting started"
+  },
+  "Once you've downloaded a Pulsar release, instructions on getting up and running with a standalone cluster that you can run on your laptop can be found in the": {
+    "message": "Once you've downloaded a Pulsar release, instructions on getting up and running with a standalone cluster that you can run on your laptop can be found in the"
+  },
+  "Run Pulsar locally": {
+    "message": "Run Pulsar locally"
+  },
+  "tutorial": {
+    "message": "tutorial"
+  },
+  "If you need to connect to an existing Pulsar cluster or instance using an officially supported client, see the client docs for these languages:": {
+    "message": "If you need to connect to an existing Pulsar cluster or instance using an officially supported client, see the client docs for these languages:"
+  },
+  "Older releases": {
+    "message": "Older releases"
+  },
+  "Pulsar Adapters": {
+    "message": "Pulsar Adapters"
+  },
+  "Pulsar Adapters are available on Maven Central, there is no binary package.": {
+    "message": "Pulsar Adapters are available on Maven Central, there is no binary package."
+  },
+  "Apache Pulsar Manager downloads": {
+    "message": "Apache Pulsar Manager downloads"
+  },
+  "Pulsar Manager older releases": {
+    "message": "Pulsar Manager older releases"
+  },
+  "Latest Stable Version": {
+    "message": "Latest Stable Version"
+  },
+  "Latest Version": {
+    "message": "Latest Version"
+  },
+  "Here you can find the latest documentation and unreleased code.": {
+    "message": "Here you can find the latest documentation and unreleased code."
+  },
+  "Past Version": {
+    "message": "Past Version"
+  },
+  "Here you can find documentation for previous versions of Apache Pulsar.": {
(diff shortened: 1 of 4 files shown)


urfreespace:
image

image

site2/website-next/src/theme/DocVersionBanner/index.js
```diff
@@ -25,7 +25,7 @@ function UnreleasedVersionLabel({ siteTitle, versionMetadata }) {
       }}
     >
       {
-        "This is unreleased documentation for {siteTitle} {versionLabel} version."
+        "This documentation is for an unreleased {siteTitle} {versionLabel} version of Apache Pulsar. "
```
momo-jun, line 28: Seems {siteTitle} is duplicate with `Apache Pulsar`, and maybe {versionLabel} can also be removed.

tisonkun:
```suggestion
        "This documentation is for an unreleased version of Apache Pulsar. "
```

Agree. Simply writing "an unreleased version" works fine for me.

site2/website-next/src/theme/DocVersionBanner/index.js
```diff
@@ -41,7 +41,7 @@ function UnmaintainedVersionLabel({ siteTitle, versionMetadata }) {
       }}
     >
       {
-        "This is documentation for {siteTitle} {versionLabel}, which is no longer actively maintained."
+        "This documentation is for an out-of-date {siteTitle} {versionLabel} version of Apache Pulsar. "
```
tisonkun, line 44:
```suggestion
        "This documentation is for an out-of-date version of Apache Pulsar. "
```

Ditto. The sidebar and URL should already identify which version it is.

tisonkun (changes requested):

Anonymitaet:
Thanks! @urfreespace 

For the banner, can we consider using the following? @tisonkun @momo-jun @DaveDuggins @D-2-Ed

🔹🔹 🔹 

For `Next` docs:

> This is unreleased documentation for Pulsar Next.
> For the latest stable version, see [2.10.x](URL-placeholder). 

🔹🔹 🔹 
 
For the latest stable doc version (eg. 2.10.x):

> This is the latest stable version [2.10.x](URL-placeholder).

🔹🔹 🔹 

For docs in the maintenance cycle (eg. 2.9.x, 2.8.x)

> This is the documentation for Pulsar 2.9.x
> For the latest stable version, see [2.10.x](URL-placeholder). 

🔹🔹 🔹 

For docs not in the maintenance cycle (eg. versions prior to 2.8.x)

> This is the documentation for Pulsar 2.6.2, which is no longer actively maintained.
> For the latest stable version, see [2.10.x](URL-placeholder).

site2/website-next/scripts/code.json
```diff
@@ -0,0 +1,501 @@
...
+  "theme.docs.versions.unreleasedVersionLabel": {
+    "message": "This documentation is for an unreleased {siteTitle} {versionLabel} version of Apache Pulsar. ",
+    "description": "The label used to tell the user that he's browsing an unreleased doc version"
+  },
+  "theme.docs.versions.unmaintainedVersionLabel": {
+    "message": "This documentation is for an out-of-date {siteTitle} {versionLabel} version of Apache Pulsar. ",
+    "description": "The label used to tell the user that he's browsing an unmaintained doc version"
+  },
+  "theme.docs.versions.latestVersionSuggestionLabel": {
+    "message": "We recommend you use the {latestVersionLink} ({versionLabel}).",
```
Anonymitaet, line 132: In technical writing:

- Avoid first-person pronouns. Details see https://docs.google.com/document/d/1lc5j4RtuLIzlEYCBo97AC8-U_3Erzs_lxpkDuseU0n4/edit#bookmark=id.gfwms5zehcak. 
- Avoid using "recommend" if possible. A phrase such as “We recommend that you take the following action” could create a potential marketing or legal problem

urfreespace:
@Anonymitaet @tisonkun Maybe we need to agree on the text and style of the banner

tisonkun:
Hi @Anonymitaet 

> For the latest stable doc version (eg. 2.10.x):
> 
> This is the latest stable version [2.10.x](https://github.com/apache/pulsar-site/pull/URL-placeholder).

I don't think we need a banner for the (expected) latest stable doc.

Agree on other wording changes.

> Blue is more appropriate than yellow/red here.
>
> Yellow / red can be used to indicate important hints, tips, guidance, restrictions, advice that might be overlooked, or the possibility of damage to software.
>
> If yellow / red is overused, they lose their significance.

Here we'd like to hint the users to use the latest stable doc. So it's not overuse from my perspective. You can get the image why it's a hint from:

* https://nightlies.apache.org/flink/flink-docs-release-1.12/
* https://docusaurus.io/docs/2.0.1

You _can_ find several projects that don't follow this way. But think of the Pulsar situation - we don't maintain quite a few old versions, and it's still under rapid development. We don't want users to use an old version and ask for bug fixes cherry-pick, but just try out the latest stable. For users already use an old version, the color is not a big deal.

Anonymitaet:
@tisonkun thanks!

Since the Pulsar website background is light blue, yellow / red is too obvious on light blue than it is on the white background (eg. Docusaurus).

image

image

tisonkun:
@Anonymitaet I agree that color is hard 🤣 The comment here is to make it relatively obvious, but choosing a nice color can be improved later. @labuladong @sepush do you have some input on which color is suitable for this purpose?

Sepush:
> @Anonymitaet I agree that color is hard 🤣 The comment here is to make it relatively obvious, but choosing a nice color can be improved later. @labuladong @Sepush do you have some input on which color is suitable for this purpose?

I recommend don't use blue as the background just use white for light mode, If you want blue theme,only apply for the `link` `border` or some others will be better

tisonkun:
> don't use blue as the background just use white for light mode

This is a pivotal change. Let's stop further discussion in this direction.

michaeljmarshall:
I like the banner for the `next` docs, thank you for proposing this change @Anonymitaet!

michaeljmarshall:
I'm not sure where to provide this feedback (so feel free to redirect me to a different channel), but I find the recent change to version numbers being 2.10.x to be confusing. I like to know that I am looking at the exact documentation for a version, since there are minor changes between patch versions that might affect documentation. Without explicit versioning, we lose important nuance.

labuladong:
> @Anonymitaet I agree that color is hard 🤣 The comment here is to make it relatively obvious, but choosing a nice color can be improved later. @labuladong @Sepush do you have some input on which color is suitable for this purpose?

I think yellow is fine, we can make it darker if light yellow is too obvious on light blue.

labuladong:
> I'm not sure where to provide this feedback (so feel free to redirect me to a different channel), but I find the recent change to version numbers being 2.10.x to be confusing. I like to know that I am looking at the exact documentation for a version, since there are minor changes between patch versions that might affect documentation. Without explicit versioning, we lose important nuance.

Hi @michaeljmarshall, could this `Other` button solve your problem? That page lists all versions:

image

momo-jun:
> I'm not sure where to provide this feedback (so feel free to redirect me to a different channel), but I find the recent change to version numbers being 2.10.x to be confusing. I like to know that I am looking at the exact documentation for a version, since there are minor changes between patch versions that might affect documentation. Without explicit versioning, we lose important nuance.

@michaeljmarshall you can refer to [PIP-190](https://github.com/apache/pulsar/issues/16637) for more details. In short, the changes between patch versions are annotated with a version-specific note like “This configuration is only available in 2.10.1 and later versions.”

tisonkun:
> That page lists all versions:

Unfortunately, it doesn't. It still lists one `2.10.x` entry for all 2.10.x series versions.

This is a separate topic, I suggest you start another thread to discuss it.

tisonkun:
@Anonymitaet @urfreespace improving color can be a continuous effort. If we now agree on the wording part. We can push forward the changes in wording.

I have one comment here: Don't show a banner for the latest stable version. Users are already on the version we'd recommend and nothing should be noticed.

Anonymitaet:
@tisonkun Thanks! 
I'd still prefer to keep the banner on the latest stable version to keep the behavior consistent, so that users may not wonder why "this page does not have such info while others pages have", then they may stop for a while to think about it. What users want s "Do not make me think, I want to skim and scan info quickly".

momo-jun:
My personal opinion on the banner:
1) Keep the current blue background color for the banner to minimize the effort. It's just a notification, not a striking/breaking change, and doesn't need highlights/cosmetics.
2) No show the banner for the latest stable version of the doc, to eliminate noise for those who're already looking at the right docs. They don't even have to think, because they are possibly unaware of the banner behavior for other versions of docs.

urfreespace:
As the new text is more explicit and precise so update the text from this Pr and keep the background color blue, are you all agree? @Anonymitaet @tisonkun @momo-jun

tisonkun (approved):

urfreespace:
I will merge it if there are no objections before 5:00 p.m @Anonymitaet @momo-jun

Anonymitaet:
@urfreespace Let's make a compromise and consensus on this:

☘️☘️☘️ Banner wording

Remove the banner for the `latest stable version`.

🔹🔹 🔹 

For `Next` docs:

> This is unreleased documentation for Next.
> For the latest stable version, see [2.10.x](URL-placeholder). 

🔹🔹 🔹 

~~For the latest stable doc version (eg. 2.10.x):~~

~~This is the latest stable version [2.10.x](URL-placeholder).~~

🔹🔹 🔹 

For docs in the maintenance cycle (eg. 2.9.x, 2.8.x):

> This is the documentation for 2.9.x
> For the latest stable version, see [2.10.x](URL-placeholder). 

🔹🔹 🔹 

For docs not in the maintenance cycle (eg. versions prior to 2.8.x):

> This is the documentation for 2.6.2, which is no longer actively maintained.
> For the latest stable version, see [2.10.x](URL-placeholder). 

☘️☘️☘️ Banner color

Blue

Anonymitaet (approved):

momo-jun:
@urfreespace @Anonymitaet can we remove the banner from this [page](https://pulsar.apache.org/release-notes/)?
image

Anonymitaet:
> @urfreespace @Anonymitaet can we remove the banner from this [page](https://pulsar.apache.org/release-notes/)? image

sure

PR #214, opened 2022-09-19
Add legacy docs for 2.6.0 to 2.8.0

SignorMercurio:


urfreespace (approved):

tisonkun:
@Anonymitaet @momo-jun what's the background here? I don't think we should use `2.6.0` and `2.7.0`; otherwise, I'll ask where are `2.6.1`, `2.7.1`. If we want to maintain config docs in minor version granularity, please use either `2.6` or `2.6.x`.

Anonymitaet:
@tisonkun thanks for your reminder!

@SignorMercurio can we make it 2.6.x, 2.7.x .... 2.10.x?

SignorMercurio:
@Anonymitaet Yes, I'll open a PR for this.

PR #227, opened 2022-09-27
Automate release table generation

SignorMercurio:
@urfreespace PTAL.

Note that `release-json-gen.sh` is already tested locally and runs well, but may fail in GitHub Actions.

Also replaced all `bdocs-tab:example_shell` with `shell` to use standard Markdown format for the new site.


urfreespace (approved):

SignorMercurio:
Forgot to mention that this PR closes pulsar#17514.

PR #235, opened 2022-09-28
remove token

urfreespace:


tisonkun:
So what happened here? Did we update `$GH_TOKEN` with a read-only token? Or just simply revert the hardcode token changes and pending to update the token with an INFRA ticket later?

PR #237, opened 2022-09-29
Fix release table versioning and style issues

SignorMercurio:
@urfreespace @Anonymitaet PTAL.

image

### DIFF
--- a/site2/tools/release-json-gen.sh
+++ b/site2/tools/release-json-gen.sh
@@ -56,14 +56,12 @@ echo $GH_TOKEN > "$TOKEN_FILE"
 gh auth login --with-token < "$TOKEN_FILE"
 rm "$TOKEN_FILE"
 
-# list all versions sorted by version number desc
-versions=$(gh release list -R apache/pulsar | awk '{print $1}' | sort -rV)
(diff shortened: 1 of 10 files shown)


tisonkun:
Hi @SignorMercurio! Is there any background about this patch? What problem does it want to fix now?

urfreespace (approved):

urfreespace:
> Hi @SignorMercurio! Is there any background about this patch? What problem does it want to fix now?

it's while background before

image

SignorMercurio:
@tisonkun Hi, the context is in a slack DM group with the two reviewer. The problems fixed are:

1. Replace the patch number with `x` for all versions above 2.8.
2. Remove releases that does not have a release note, or belongs to "Previous Releases".
3. Use transparent background that changes with day/night mode, instead of the white background.
4. Adjust table style as there are previously some unexpected styles.
5. Create component in `site2/website-next/src/components` and import it in `site2/website-next/release-notes/all.md` for better maintainability.

tisonkun:
Thanks for your information! DMs are OK for quickly syncing up. But I'd suggest trying to bring more context when you start a process in the community instead of letting other participants to guess :)

PR #246, opened 2022-10-10
Add missing release notes and the version column

SignorMercurio:
This PR add missing release notes for legacy tag releases:

- 2.5.0
- 2.6.3
- 2.9.1

And also add a version column. Note that `$DATA/release-*.js` are removed as they should be generated during build.

image

PR #247, opened 2022-10-11
Restore release-*.js

SignorMercurio:
#246 removed `release-*.js`, which caused build failed. This PR restores these files, but they should be overridden during each website build process.


tisonkun (approved):
Although, must we update these flies on every new releases created? Or if we can just push a dummy file with comments indicate that the content will be generated when building?

Still, I found the process we generate versioned doc hard to understand - actually these files are related to the main site instead of the doc pages, but we failed to generate versioned doc due to missing these files...

SignorMercurio:
@tisonkun We can just use a dummy file here. These files no longer need to be updated.

PR #251, opened 2022-10-13
correct homepage link and remove dangling bottom block

tisonkun:
cc @Anonymitaet @urfreespace

### DIFF
--- a/site2/website-next/src/pages/index.js
+++ b/site2/website-next/src/pages/index.js
@@ -180,7 +180,7 @@ export default function Home() {
             <PillButton variant="" target="" href={docUrl("concepts-overview")}>
               Learn More
             </PillButton>
-            <PillButton variant="grey" href={docUrl("getting-started-home")}>
+            <PillButton variant="grey" href={docUrl("getting-started-standalone")}>
               Quickstart
             </PillButton>
           </div>
--- a/site2/website-next/src/pages/index.js
+++ b/site2/website-next/src/pages/index.js
@@ -6,8 +6,7 @@ import HomeQuotes from "../components/HomeQuotes";
 import SubHeroBlock from "../components/SubHeroBlock";
 import PromoCallout from "../components/PromoCallout";
 import PillButton from "../components/PillButton";
-import GitHubIcon from "@mui/icons-material/GitHub";
-import { docUrl, githubUrl } from "../utils/index";
+
 var startWaves = function () {
   if (!SineWaves) {
     return;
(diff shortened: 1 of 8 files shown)


site2/website-next/src/pages/index.js
```diff
@@ -177,14 +177,10 @@ export default function Home() {
...
-            <PillButton
-              variant=""
-              target=""
-              href={docUrl("next/concepts-overview")}
-            >
+            <PillButton variant="" target="" href={docUrl("concepts-overview")}>
               Learn More
             </PillButton>
-            <PillButton variant="grey" href={docUrl("next/")}>
+            <PillButton variant="grey" href={docUrl("getting-started-home")}>
```
tisonkun, line 183:
```suggestion
            <PillButton variant="grey" href={docUrl("getting-started-standalone")}>
```

After the latest stable version becomes 2.11.x, we can revert this change.

momo-jun (approved):

Anonymitaet:
@tisonkun thanks for your contribution!

Are you changing the default page from image 1 to image 2?

Image 1
image

Image 2
image

Navigating uses to image1 is an intended behavior from doc's team since we want users to choose their path based on roles and objectives rather than give a "fixed" choice for them. 

Similar case:
https://docs.starburst.io/get-started/index.html
image

@D-2-Ed @DaveDuggins thoughts?

momo-jun:
If I understand the PR correctly, it changes the redirection of the `QUICKSTART` button on the home page from Image1 to Image2. Because the `Get Started > Quickstart` button on the top also redirects to Image2. It's quite confusing when the buttons with the identical label redirect to two different pages. That's how I understand the issue.

image

tisonkun:
@momo-jun is correct, while the original purpose is that we just do not guide people to the "next" version of docs.

@Anonymitaet whether to "about" or to "getting-started-home" is OK. If you agree, I can change both buttons to "about" and push forward this patch.

Anonymitaet:
@momo-jun @tisonkun Oh I see, thanks for your explanations!

Navigating users to the "About" page makes sense if users click any of these:
image

Anonymitaet (approved):

tisonkun:
Updated. You can preview with: `./preview.sh 2.10.x`

tisonkun:
@Anonymitaet @momo-jun shall we merge this patch now?

momo-jun:
@tisonkun I requested the review from @urfreespace since the changed tech details under `src` in the last commit need more eyes.

urfreespace (approved):

PR #252, opened 2022-10-13
fix blog edit url

tisonkun:
blog sources have been changed.

cc @Anonymitaet @urfreespace


Anonymitaet (approved):

site2/website-next/docusaurus.config.js
```diff
@@ -380,7 +381,7 @@ module.exports = {
         },
         blog: {
           showReadingTime: true,
-          editUrl: `${githubUrl}/edit/master/site2/website`,
+          editUrl: `${githubSiteUrl}/edit/main/site2/website-next/`,
```
urfreespace, line 384: I don't think we should edit docs in repo `pulsar-site`, as you know, we sync docs from main repo `pulsar`,so the changes of docs you edit directly in `pulsar-site` will be overwritten by the sync script @tisonkun

urfreespace (changes requested):

urfreespace (approved):

PR #255, opened 2022-10-18
[fix] Concept link jump to the default version of doc

labuladong:
The `Pulsar Concepts` jumps to the next version document:

image

But it should jump to the default version document as other buttons do.

### DIFF
--- a/site2/website-next/docusaurus.config.js
+++ b/site2/website-next/docusaurus.config.js
@@ -162,8 +162,7 @@ module.exports = {
           position: "right",
           items: [
             {
-              type: "doc",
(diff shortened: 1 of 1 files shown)


labuladong:
cc @urfreespace

tisonkun:
@labuladong do you clear the cookie and cache? If you try in a Incognito window, it should guide you to the latest stable version.

labuladong:
Weird, it works correctly in an incognito window. I will close this pr.

tisonkun:
@labuladong Yep..I think it's about cookie or browser cache.

PR #259, opened 2022-10-20
Improve TOC style, docs for CLI tools and directory structure

SignorMercurio:
1. Show entire left TOC, and make it collapsible
2. Fix 404 issue of CLI tools page
3. Unify behavior of docs for CLI tools
4. Remove broker-tool and add pulsar-shell
5. Remove `_sidebar.md` in each directory and add it in each version's directory
6. Move `$WEBSITE/docsify` to `$WEBSITE/static/reference` so that we can preview it together with the main site with `yarn start`
7. Add missing release notes
8. Move docs for CLI tools to this repo
9. Minor fixes and rewords


tisonkun:
I notice that after #261 merged, static files under `$WEBSITE/static/reference` aren't released. It's weird and should not happen. Otherwise, that we're moving the docsify files here becomes meaningless.

cc @urfreespace any ideas?

urfreespace:
> I notice that after #261 merged, static files under `$WEBSITE/static/reference` aren't released (for example, the `.htaccess` file). It's weird and should not happen. Otherwise, that we're moving the docsify files here becomes meaningless.
> 
> cc @urfreespace any ideas?
> 
> UPDATE: Well. I notice that only part of files under `$WEBSITE/static` are manually moved. Then it's not an issue for this patch.

fixed with https://github.com/apache/pulsar-site/pull/263

site2/website-next/release-notes/versioned/pulsar-cs-2.4.1.md
```diff
@@ -0,0 +1,9 @@
+---
```
tisonkun, line 1: Why do we need to make changes on these files (under `release-notes/versioned`.

SignorMercurio: These are newly added, not modified. They are added because they are missing.

tisonkun: Is their missing related to this patch? I don't know why they're missed in the first place and why we couple these changes in the patch to resolve the location of reference docs.

SignorMercurio: No, I also have no idea why these docs are missing. I added these to avoid 404 links in the Release Notes page.

urfreespace (approved):

site2/tools/build-site.sh
```diff
@@ -50,11 +50,8 @@ bash scripts/split-version-build.sh $@
...
 
 CONTENT_DIR="$GEN_SITE_DIR"/content
-
-rm -rf "$CONTENT_DIR"
-mkdir -p "$CONTENT_DIR"
-mkdir -p "$CONTENT_DIR"/reference
-rsync -a ./docsify/ "$CONTENT_DIR"/reference
+rm -rf "$CONTENT_DIR" && mkdir -p "$CONTENT_DIR"
+cp -R static/reference "$CONTENT_DIR"
 rsync -a ./build/ "$CONTENT_DIR"
```
tisonkun, line 55: `cp -R` seems work differently to the standard on the CI env. Use `rsync -a` correctly stably copy all files (including hidden files) recursively.

Besides, since we move references under the `static` folder, there's no need to special handle them.

Thus, this code snippet can be written as:

```bash
rm -rf "$CONTENT_DIR"
mkdir -p "$CONTENT_DIR"
rsync -a ./build/ "$CONTENT_DIR"
```

@urfreespace @SignorMercurio please check.

tisonkun (changes requested):
I think one of the object we make this change is simplifying the build stage. Comments inline.

site2/website-next/release-notes/versioned/pulsar-cs-2.4.1.md
tisonkun: So where do you add these files from?

site2/tools/build-site.sh
urfreespace: yes, thanks for @tisonkun, could you pls do a test with this code from @tisonkun's suggestion? @SignorMercurio

Anonymitaet:
LGTM from the requirement side.

site2/tools/build-site.sh
SignorMercurio: Sure, I'll have a try and update the script.

tisonkun:
I just noticed that we should still generate CLI references for each major release. Is it correct?

PR #263, opened 2022-10-24
Fix the bug that the redirection file is not copied when building

urfreespace:


site2/website-next/scripts/split-version-build.sh
```diff
@@ -76,5 +76,6 @@ done <scripts/.versions
 
 cp -r build-assets/* build/assets/
 rm -rf build-assets
+cp static/.htaccess build/
```
tisonkun, line 79: Shall we copy all static files to build here and specially handle the hidden files like `.htaccess`?

In this way, we don't have to specially copy static files in:

https://github.com/apache/pulsar-site/blob/c30db65192c9f1791bf4bd2cf1b300124643933a/site2/tools/build-site.sh#L45-L62

tisonkun (commented):
Thanks for taking care of this issue! Looks good to me.

Comment inline. And you can write `close apache/pulsar#18174` instead of just referring it.

tisonkun (changes requested):
It seems that if we simply write:

```bash
cp -R ./build/ "$CONTENT_DIR"
```

We can copy the hidden files as well. No need for special handling.

https://superuser.com/questions/61611/how-to-copy-with-cp-to-include-hidden-files-and-hidden-directories-and-their-con

site2/website-next/scripts/split-version-build.sh
urfreespace: good idea

urfreespace:
fixed, PTAL again, thanks @tisonkun

site2/website-next/scripts/split-version-build.sh
tisonkun: It seems that we don't need this line then.

site2/tools/publish-website.sh
```diff
@@ -65,7 +65,7 @@ PULSAR_SITE_TMP=/tmp/pulsar-site
       | grep -v $PULSAR_SITE_TMP/content/.htaccess \
       | xargs rm -rf
   fi
-  cp -r $GENERATED_SITE_DIR/content/* $PULSAR_SITE_TMP/content
+  cp -r $GENERATED_SITE_DIR/content/ $PULSAR_SITE_TMP/content
```
tisonkun, line 68: Perhaps using `-R` to avoid behavior inconsistency. For the implementation on a darwin OS:

```
     Historic versions of the cp utility had a -r option.  This implementation
     supports that option, however, its behavior is different from historical
     FreeBSD behavior.  Use of this option is strongly discouraged as the
     behavior is implementation-dependent.  In FreeBSD, -r is a synonym for -RL
     and works the same unless modified by other flags.  Historical
     implementations of -r differ as they copy special files as normal files
     while recreating a hierarchy.
```

site2/tools/build-site.sh
```diff
@@ -58,7 +58,7 @@ rm -rf "$CONTENT_DIR"
 mkdir -p "$CONTENT_DIR"
 cp -R "$GEN_SITE_DIR"/reference "$CONTENT_DIR"
 # cp -R "$GEN_SITE_DIR"/api "$CONTENT_DIR"
-cp -R ./build/* "$CONTENT_DIR"
+cp -R ./build/ "$CONTENT_DIR"
 cp -R "$WEBSITE_DIR"/static/swagger/* "$CONTENT_DIR"/swagger/
```
tisonkun, line 62: Perhaps we don't need this line since the static swagger is copyed at the line above.

tisonkun (commented):
Comments inline.

site2/tools/build-site.sh
tisonkun: And we can add a `ls` command to debug why https://github.com/apache/pulsar-site/actions/runs/3317429325/jobs/5480293467 failed.

site2/tools/publish-website.sh
urfreespace: fixed with https://github.com/apache/pulsar-site/pull/264

site2/tools/build-site.sh
urfreespace: fixed with https://github.com/apache/pulsar-site/pull/264

PR #264, opened 2022-10-25
fix build error caused by cp swagger dir

urfreespace:
fixed build error, detail https://github.com/apache/pulsar-site/actions/runs/3317429325/jobs/5480293467

### DIFF
--- a/site2/tools/build-site.sh
+++ b/site2/tools/build-site.sh
@@ -50,7 +50,7 @@ bash scripts/split-version-build.sh $@
 cd "$WEBSITE_DIR"
 
 mkdir -p "$GEN_SITE_DIR"/reference
-cp -r docsify/* "$GEN_SITE_DIR"/reference
+cp -R docsify/* "$GEN_SITE_DIR"/reference
 
 CONTENT_DIR="$GEN_SITE_DIR"/content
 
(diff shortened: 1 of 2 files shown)


urfreespace:
@tisonkun PTAL, thanks

tisonkun (approved):

tisonkun:
It seems the result https://github.com/apache/pulsar-site/commit/f17fd0698c2400bb69cef8aeba20c5f49b289e15 is under `content/content/build` which has two more level inner then the expected one.

site2/tools/publish-website.sh
```diff
@@ -65,7 +65,7 @@ PULSAR_SITE_TMP=/tmp/pulsar-site
       | grep -v $PULSAR_SITE_TMP/content/.htaccess \
       | xargs rm -rf
   fi
-  cp -r $GENERATED_SITE_DIR/content/ $PULSAR_SITE_TMP/content
+  cp -R $GENERATED_SITE_DIR/content/ $PULSAR_SITE_TMP/content
```
tisonkun, line 68: The root cause should be that we change from `-r` to `-R` and the behavior changed.

site2/tools/build-site.sh
```diff
@@ -50,7 +50,7 @@ bash scripts/split-version-build.sh $@
 cd "$WEBSITE_DIR"
 
 mkdir -p "$GEN_SITE_DIR"/reference
-cp -r docsify/* "$GEN_SITE_DIR"/reference
+cp -R docsify/* "$GEN_SITE_DIR"/reference
```
tisonkun, line 53: And thus we should also take care of this line.

PR #277, opened 2022-11-09
Remove disableSwitch

zeo1995:


tuteng (approved):

tisonkun:
@zeo1995 @tuteng could you write a bit description about this change? What's the problem (motivation) and how this patch resolves the problem.

urfreespace:
https://www.notion.so/streamnativeio/fa562f704de941909733d6927a290b93?v=d3d6ca95618d4c4cb7c16608e104330d&p=d8a145b189d9464ea0cf15b6c8c47cfe&pm=s 
@tisonkun

tisonkun:
OK. Now I learn the purpose from the PR title. But I think if a community member isn't member of streamnativeio, he/she cannot read the note you share :)

dave2wave:
I like the dark mode option. It is not acceptable to discuss project decisions on a place that is not visible to the whole community.

dave2wave:
I would like for this PR to be reverted.

tuteng:
Sorry for my wrong merge, I revert this change now, please have a look at it, these should need to be discussed in more detail @dave2wave @tisonkun

merlimat:
@dave2wave You're right that the link to the internal discussion is meaningless and this should be discussed here. 

To add a bit of my context on why this should be removed (in its current state), the main issue is that there are contrast and accessibility issues in dark mode: 

 * Diagrams with white background look very bad
 * Diagrams with transparent background look unreadable
 * Menu blue doesn't have contrast
 * Dropdown menu options are not visible
 * Logo is not readable
 * Alert banners don't work in a dark background
 * Problems with contrasting colors
 * Case studies page - not readable
 * There are a lot of these issues scattered across the website
 
It would take a huge of effort to make sure the documentation pages look decent in both color schemes. Right now we should at least make it look decent in 1 scheme first :).

tisonkun:
@merlimat I agree with your statement, especially on the style side. IMO the font size and font family and line spacing are all suboptimal now. But yes, it would take a huge of effort to improve. Both the light mode and the dark mode. And the dark mode currently more lacks maintenance and has extra issues as you stated.

PR #281, opened 2022-11-10
fix redirections with/without trailing slash

tisonkun:
This patch is based on #280 (There is no logical dependency, but my local env :)).

This closes https://github.com/apache/pulsar/issues/18410.


tisonkun:
Well. This patch should be picked to the upstream also.

urfreespace (approved):

PR #283, opened 2022-11-11
Document how to generate apidoc python

tisonkun:
This closes https://github.com/apache/pulsar/issues/18347.

cc @michaeljmarshall please take a look.

### DIFF
--- a/site2/tools/apidoc/python/pydoctor-generator-helper.sh
+++ b/site2/tools/apidoc/python/pydoctor-generator-helper.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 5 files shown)


site2/tools/apidoc/python/pydoctor-generator-helper.sh
```diff
@@ -0,0 +1,38 @@
...
+
+if [ -z "$PULSAR_VERSION" ]; then
+    echo "PULSAR_VERSION must be set."
+    exit 1
+fi
+
+python -m pip install --upgrade pip
+
+pip install -U pydoctor
+git clone https://github.com/apache/pulsar-client-python --single-branch --branch v$PULSAR_VERSION
```
tisonkun, line 32: cc @merlimat @BewareMyPower @RobertIndie this line depends on that for each Python client release, we create a `vX.Y.Z` tag. I notice that C++ client does it so make this assumption.

tisonkun: Verified the overall script by hardcoding the tag as `main` locally.

site2/tools/apidoc/python/pydoctor-generator.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
```
tisonkun, line 1: I know that these two script files duplicate each other. But since scripts are less combinable, and this is a specific small scope, I'd like to keep it simple stupid :)

michaeljmarshall (approved):
Thanks for updating this @tisonkun!

RobertIndie (approved):

tisonkun:
Merging...

If later we have different release process for Python client, we can update it correspondingly.

PR #288, opened 2022-11-15
Add alias for latest version

tisonkun:
This closes https://github.com/apache/pulsar/issues/17438.

The `client-redirects` plugin [only works in production](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects), so you should preview with `yarn build` and `yarn serve` to verify that `/docs/` can be redirected to `/docs/{latestVersion}`. I made some changes to use concrete versions when possible.


urfreespace:
@tisonkun in fact, you could mock the production model by run `docusaurus build && docker-compose up`

urfreespace (approved):
LGTM

tisonkun:
Merging...

PR #289, opened 2022-11-15
Cleanup legacy scripts to generate API docs

tisonkun:
Signed-off-by: tison <wander4096@gmail.com>


site2/tools/generate-api-docs.sh
```diff
@@ -1,37 +0,0 @@
...
-cd $SCRIPT_DIR
-
-ROOT_DIR=$(git rev-parse --show-toplevel)
-VERSION=`${ROOT_DIR}/src/get-project-version.py`
-
-set -x -e
-
-cd ${ROOT_DIR}
-mkdir -p site2/website/static/swagger/${VERSION}
-cp pulsar-broker/target/docs/*.json site2/website/static/swagger/${VERSION}/
```
tisonkun, line 31: Although we copy `2.11.0-SNAPSHOT` alike files here, it's not reflected in the finally result. I suppose it gets filtered by `scripts/split-swagger-by-version.js` so that this whole file is useless to be removed.

urfreespace (approved):

site2/tools/generate-api-docs.sh
urfreespace: yeah, you are right

PR #290, opened 2022-11-15
Remove logics to replace cpp / python apidoc

tisonkun:
Now they have their own repository and release cycle, it's impossible to infer the proper version from the current Pulsar version. We should instead hardcode the corresponding versions.

This patch should be merged after we change the source links in the main repo.

Signed-off-by: tison <wander4096@gmail.com>


tisonkun:
I have another solution now: we can use a `@pulsar:apidoc_python@`/`@pulsar:apidoc_cpp@` for legacy versions, and use the new hard link reference for new versions.

PR #291, opened 2022-11-15
Specially handle Python and C++ API documents 

tisonkun:
Now they have their own repository and release cycle, it's impossible to infer the proper version from the current Pulsar version. We should instead hardcode the corresponding versions for new versions.

This patch should be merged after we change the source links in the main repo https://github.com/apache/pulsar/pull/18477.


tisonkun:
https://github.com/apache/pulsar/pull/18477 merged. Merging...

PR #292, opened 2022-11-15
Remove legacy CLI pages

tisonkun:
Now we release CLI pages at https://pulsar.apache.org/reference/.

Legacy pages are released under referece-xxx or /tools, these removed pages are not referred in search engines so I tend not to maintain multiple pages for the same purpose.


tisonkun:
Merging...

If it causes new broken links, please share the real-world cases and I'll investigate alternative solutions.

PR #306, opened 2022-11-28
Fix gpg command in validate-release-candidate.md

RobertIndie:
The gpg command example in `validate-release-candidate` is wrong. This PR fixes it.


tisonkun (approved):
Thank you! Merging...

PR #318, opened 2022-12-06
Improve bookkeeper server status metrics

codelipenghui:
Add missed metrics for server status `-1`
Details to see 
https://github.com/apache/bookkeeper/blob/4ee8c0411878689417c39b7027899ed7ce815e7e/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java#L87-L90

### DIFF
--- a/site2/website-next/docs/reference-metrics.md
+++ b/site2/website-next/docs/reference-metrics.md
@@ -38,7 +38,7 @@ in the `bookkeeper.conf` configuration file.
 
 | Name | Type | Description |
 |---|---|---|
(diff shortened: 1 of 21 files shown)


hangc0276 (approved):

tisonkun (changes requested):
You may need to update the source files like https://github.com/apache/pulsar/blob/master/site2/docs/reference-metrics.md.

Closing...

I'm thinking of moving all docs sources to the site repo, but it can take appreciable efforts to implement.

PR #327, opened 2022-12-13
[improve][doc] add a tip for skipping test procedure for md contribution

Anonymitaet:
Preview looks good:
image


tisonkun (changes requested):
This is redundant and the tests for docs-only change will be passed in seconds. We don't need to document this.

PR #335, opened 2022-12-14
Change link to Slack workspace auto-invite

merlimat:
The Heroku app that was being used to automate the Slack invite of new users has stopped working since Heroku has terminated the free tier.

I've added a new Slack app that does the email invite.


tisonkun (approved):
Thanks for your updates. I check the link and it looks good (without actually signing).

Merging...

PR #352, opened 2022-12-31
Release 2.9.4

tisonkun:
cc @congbobo184 

cc @BewareMyPower Do we release C++/Python client for 2.9.4 (and following 2.8.x, 2.9.x, 2.10.x versions)?

I'm afraid that some of the doc references to the latest client version will be resolved as 2.9.4, but we don't actually release such clients and cause issues.


tisonkun:
@michaeljmarshall @tuteng I don't think we should generate swagger files for every minor release. It's the same as API docs. But current REST API pages and some version resolving code depend on this XD.

BewareMyPower:
No. We won't release C++/Python clients for 2.x.y.

congbobo184 (approved):

tisonkun:
@congbobo184 Thanks for your help! Merging...

@BewareMyPower Thanks for your information. Then I'm going to change the client libraries pages to fix on the latest release version later.

PR #355, opened 2023-01-01
Add Scala Neutron library from ProfunKtor

gvolpe:


tisonkun (commented):
@gvolpe Interesting.

May I ask what's the relation between https://github.com/cr-org/neutron and https://github.com/profunktor/neutron? It seems you're the author for both projects. If so, we may replace the previous lib with the new (maintained?) one.

gvolpe:
@tisonkun fair question 😄 More details here: https://github.com/profunktor/neutron#disclaimer

I started out `cr-org/neutron` when I was working at that company, but I forked it to add more features, documentation and to support Scala 3, so it has now deviated quite a bit from the other one.

I know they are still using it in production, but can't speak up for that project anymore.

tisonkun:
@gvolpe Thanks for your information.

Then I'm going to merge these patches plus mark your fork as `(profunktor's fork)`.

The ASF discourages "hostile fork," and the Apache License 2.0 doesn't grant trademarks to the recipients. I'll suggest you choose a different name or try to negotiate with the cr-org team to find a solution. Otherwise, it's at risk of trademark issues.

I'll assume good motivations as you maintain a client library of Apache Pulsar. And if cr-org people don't complain about the project name confusion, it should be fine.

cc @merlimat @dave2wave what do you think?

gvolpe:
That's fine, I don't think having the same name is an issue. We talked about unifying it at some point in favor of the ProfunKtor's fork, but it depends on cr-org adopting this client, which requires work that probably lies at the end of their backlog.

If there's any issue at some point, I'm willing to collaborate to find a resolution.

tisonkun:
Merging...

A follow-up will be landed soon to add the `(profunktor's fork)` marker.

PR #364, opened 2023-01-08
[fix][doc] Correct configuration requirements for the HDFS sinks

sekikn:
[The HDFS2 sink documentation refers to the `filenamePrefix` configuration as "(Required) true, if compression is set to None"](https://pulsar.apache.org/docs/next/io-hdfs2-sink/#property), but [it's always required actually. Instead, either `fileExtension` or `compression` should be set at least](https://github.com/apache/pulsar/blob/master/pulsar-io/hdfs2/src/main/java/org/apache/pulsar/io/hdfs2/sink/HdfsSinkConfig.java#L96-L99).
[This behaviour is the same for HDFS3](https://github.com/apache/pulsar/blob/master/pulsar-io/hdfs3/src/main/java/org/apache/pulsar/io/hdfs3/sink/HdfsSinkConfig.java#L86-L89
), so its documentation also should be fixed.


Anonymitaet:
Ping @Demogorgon314 again for technical review, thanks!

Demogorgon314 (approved):
LGTM.

Anonymitaet (approved):

PR #365, opened 2023-01-08
[fix][doc] Fix JDK version for release process

sekikn:
[The release process document says JDK 11 is required](https://pulsar.apache.org/contribute/release-process/#preparation), but 17 is supposed to be required for recent releases actually.

### DIFF
--- a/contribute/release-process.md
+++ b/contribute/release-process.md
@@ -34,7 +34,7 @@ If you haven't already done it, [create and publish the GPG key](create-gpg-keys
 
 Before you start the next release steps, make sure you have installed these software:
 
-* JDK 17
+* JDK 11 (up to 2.10.x) or 17 (2.11+)
 * Maven 3.8.6
 * Zip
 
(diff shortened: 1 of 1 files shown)


tisonkun (changes requested):
Thanks for your contribution @sekikn!

Since the current maintained versions is from 2.8~master, I'm afraid that for 2.8~2.10 we're still requiring JDK 11, while from 2.11 we require JDK17. Can you update the description to reflect this case?

sekikn:
Thank you for your comment @tisonkun! Just update the PR.

contribute/release-process.md
```diff
@@ -34,7 +34,7 @@ If you haven't already done it, [create and publish the GPG key](create-gpg-keys
 
 Before you start the next release steps, make sure you have installed these software:
 
-* JDK 11
+* JDK 11 (up to 2.10.x) or 17 (2.11+)
```
tisonkun, line 37:
```suggestion
* JDK 17 (for Pulsar version >= 2.11) or JDK 11 (for earlier versions)
```

tisonkun (approved):
LGTM. Merging...

PR #368, opened 2023-01-09
ci: warn on broken links

tisonkun:
* This closes https://github.com/apache/pulsar/issues/18026.
* This refers to https://github.com/apache/pulsar/discussions/19144

I'm going to integrate the build stage in the pre-commit workflow then we can have the broken links detection in CI.

Current output for broken markdown links (partial):

image

... and broken links:

image

Some are missing pages; others are markers to be replaced on build. We should sort them out and conquer separately. Most markers to be replaced on build should be able to replace with TSX components.

cc @momo-jun @Anonymitaet @urfreespace


Technoboy- (approved):

momo-jun:
@tisonkun Thanks for enabling this feature!
Since I've been fixing the reported broken links through https://github.com/apache/pulsar-site/pull/406, I will double-check and fix all the broken links that have been warned during the local build for 2.8.x and later versions.

PR #374, opened 2023-01-11
[build][pytools] Build site and precommit

tisonkun:
* This closes https://github.com/apache/pulsar/issues/19186.
* This closes https://github.com/apache/pulsar/issues/19185.

There are still some issues with catching changed files, but it's inherited from the original script. Maybe revisit later. (And we can always build all)


tisonkun:
CI preview looks good:

image

image

PR #388, opened 2023-01-23
Update io-debezium-source.md

Sumitkumar1331:
Inputs field will not work in setting up debezium connector. You will need topicName instead of it


tisonkun (changes requested):
@Sumitkumar1331 Thanks for your contribution! Could you try to investigate what maintained versions (2.8~2.11) this issue affects?

And you can update all of those versions doc and also the next version under `docs/io-debezium-source.md`

tisonkun:
Close as no response.

PR #389, opened 2023-01-26
version mismatch

medsandor:
changed to 2.11.0 from 2.10.0 in the example

example won't work with 2.10.0, but works with 2.11.0


tisonkun:
Although I may prefer `@pulsar:version@` here, this patch is valuable. Also I'm trying to get rid of in-place text replacement.

PR #392, opened 2023-01-30
[feat] add document feedback button

labuladong:
Add feedback component on the site:

image

The feedback (up or down) of users will be converted to matomo event:

image

Then we can analyze the feedback on matomo https://matomo.privacy.apache.org/index.php?module=CoreHome&action=index&date=yesterday&period=day&idSite=32#?period=year&date=2023-01-01&idSite=32&category=General_Actions&subcategory=Events_Events

Below is my test data:

image


momo-jun:
@labuladong, thanks for adding this feature! It's extremely helpful to gather user feedback. I also have a couple of questions:
- Is it a global setting to enable this button on every doc page, or do we need to do some manual additions for each one?
- Is it possible to pop up a dialog box to request further comments after users click the down button? This can help address the root cause and also avoid random clicks. Here is an example:
image

labuladong:
1. It's a global setting. Actually, it is a component that floats on Toc. If a page has a Toc, then this component will appear on it.
2. I think it's possible to implement. Can you give me some example sites?

Anonymitaet:
@labuladong 

1. Does it make sense to move the feedback part to the end of a page or the end of the right TOC? Sitting on top of the right TOC attracts much attention, which should be left for the "core content".

Example: https://developer.hashicorp.com/terraform/tutorials/aws-get-started/infrastructure-as-code?in=terraform%2Faws-get-started
image

2. Does it make sense to add this feature to other pages on the Pulsar website? 
e.g., [contribution page](https://pulsar.apache.org/contribute/), blog page, etc.

4. Does `val` mean `value`? I searched it on google but got no accurate results.

image

image

labuladong:
1. I think the position is fine. Users can give feedback not only at the bottom of pages.
2. It will appear on all pages if this page has TOC.
3. `val` is my test data. The normal data is `up` and `down` for now.
@Anonymitaet

tisonkun (commented):
May you try to write in typescript for code intelligence and maintenance?

labuladong:
Now all code is JSX, so I think it's not necessary to use TS in this component. If you want to use TS, maybe we can start a new project to rewrite all JSX code sometime. @tisonkun

momo-jun:
> I think it's possible to implement. Can you give me some example sites?

Here is an example from [Redpanda docs](https://docs.redpanda.com/docs/get-started/intro-to-events/). The feedback buttons appear both following the TOC and at the bottom of a doc page. Only the down button triggers the dialog box.

image

urfreespace (approved):
Good job! I think we could accept this first as it matches the goal of the feedback function, we could make another PR to optimize the experience about feedback, what do you think? @momo-jun @Anonymitaet @tisonkun

tisonkun (approved):
Good to go :)

labuladong:
Agree, let's improve this in further PRs if needed.

tisonkun:
Merging..

Thanks for your contribution @labuladong!

PR #405, opened 2023-02-10
Fix set-persistent admin doc

tisonkun:
This supersedes https://github.com/apache/pulsar/pull/19001.

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/docs/admin-api-namespaces.md
+++ b/docs/admin-api-namespaces.md
@@ -385,13 +385,13 @@ admin.namespaces().removeBacklogQuota(namespace, backlogQuotaType)
 
 Persistence policies allow users to configure persistency-level for all topic messages under a given namespace.
 
-  -   Bookkeeper-ack-quorum: Number of acks (guaranteed copies) to wait for each entry, default: 0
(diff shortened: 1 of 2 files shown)


Anonymitaet:
@tjiuming PTAL, thanks!

tisonkun:
One important issue that what versions are affected. From https://github.com/apache/pulsar/pull/18999 I can see only the NEXT version (3.0.0) is affected. Please confirm.

tisonkun:
@Anonymitaet Thanks for your information.

Let's think of the change: the change commit to doc site will be published immediately, while the patch release will be out weeks or months later. It may not be a good idea to include changes to those versions now, but it helps we share information here.

Anonymitaet:
We can tag this PR with the same labels and add docs to 2.9.x, 2.10.x, 2.11.x after those versions are released.

tjiuming:
The code PR tagged with 2.9 2.10 2.11, I think keep align with it will be fine

tisonkun:
Updated. Merging...

PR #407, opened 2023-02-13
[improve][doc] Add note on client/broker compatibility

onobc:
### Documentation

Adds a note to the client overview page that mentions the client/broker compatibility.

#### Screenshot

Screen Shot 2023-02-13 at 14 58 46

- [x] `doc` 

@michaeljmarshall @lhotari @david-streamlio can you take a peek?

### DIFF
--- a/docs/client-libraries.md
+++ b/docs/client-libraries.md
@@ -26,7 +26,7 @@ Pulsar supports the following language-agnostic client libraries:
 
 **Client / Broker compatibility**
 
-A design goal of Pulsar is to ensure full compatibility between the same major version clients and brokers, e.g. 2.x clients are compatible with all 2.x brokers, etc.
+A design goal of Pulsar is to ensure full compatibility between all versions of the client and the broker. When a client connects to a broker they agree upon a version of the protocol to use. As a result, new features that rely on an updates to the protocol are only available when using both newer clients and newer brokers.
 
 :::
 
--- a/docs/client-libraries.md
+++ b/docs/client-libraries.md
@@ -22,6 +22,14 @@ Pulsar supports the following language-agnostic client libraries:
 | REST      | [User doc](client-libraries-rest.md)      | [Bundled](pathname:///release-notes/)             | [Bundled](https://github.com/apache/pulsar/tree/master/pulsar-broker)    |
 | WebSocket | [User doc](client-libraries-websocket.md) | [Standalone](pathname:///release-notes/client-ws) | [Bundled](https://github.com/apache/pulsar/tree/master/pulsar-websocket) |
 
+:::note 
+
+**Client / Broker compatibility**
+
+A design goal of Pulsar is to ensure full compatibility between the same major version clients and brokers, e.g. 2.x clients are compatible with all 2.x brokers, etc.
+
+:::
+
 ## Feature matrix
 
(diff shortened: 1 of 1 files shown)


docs/client-libraries.md
```diff
@@ -22,6 +22,14 @@ Pulsar supports the following language-agnostic client libraries:
 | REST      | [User doc](client-libraries-rest.md)      | [Bundled](pathname:///release-notes/)             | [Bundled](https://github.com/apache/pulsar/tree/master/pulsar-broker)    |
 | WebSocket | [User doc](client-libraries-websocket.md) | [Standalone](pathname:///release-notes/client-ws) | [Bundled](https://github.com/apache/pulsar/tree/master/pulsar-websocket) |
 
+:::note 
+
+**Client / Broker compatibility**
+
+A design goal of Pulsar is to ensure full compatibility between the same major version clients and brokers, e.g. 2.x clients are compatible with all 2.x brokers, etc.
```
onobc, line 29: I reached out and chatted w/ a few committers on client/broker compatibility as I found no mention of it in any docs. 

This is the general summarized consensus.

michaeljmarshall: By my understanding, the goal is broader. The Pulsar binary protocol has never introduced breaking changes. The project's goal is to ensure full compatibility between brokers and clients across all versions. This is especially important since we're using a bit of a different understanding of major versions, and we're likely going to release 3.0 soon, and there is no foreseeable reason the protocol will break.

For reference, we version the protocol by adding a new `ProtocolVersion` enum value any time a feature is added. When the client and broker connect, they establish a version to use, and both are supposed to adapt their behavior depending on the agreed upon version.

Here is the source for that enum: https://github.com/apache/pulsar/blob/950ff441da28e144bdfb71c317a9bc339d4f05b7/pulsar-common/src/main/proto/PulsarApi.proto#L241-L266

@merlimat would you be able to speak to the historical evolution of the protocol?

michaeljmarshall (commented):
Great start @onobc. This will be a great addition to the client documentation.

onobc:
Thanks @michaeljmarshall for that detail and the links to the source. 

Maybe we should keep a "lightweight" note here in the client section that quickly answers "What client version can I use w/ my broker version?". 
Then that note can link to more details that talks about the overall design goal, the protocol, handshake etc..

Wdyt?

tisonkun:
> Maybe we should keep a "lightweight" note here in the client section that quickly answers, "What client version can I use w/ my broker version?".

Since we have the `ProtocolVersion` to determine compatible feature sets, I think the answer is "any client >= the broker version".

The note about semantic version is not quite matching since we now use the "major version" as a notation of "LTS version" - https://pulsar.apache.org/contribute/release-policy/

docs/client-libraries.md
michaeljmarshall, line 29: Sorry for the delay here @onobc. What do you think about the following?

```suggestion
A design goal of Pulsar is to ensure full compatibility between all versions of the client and the broker. When a client connects to a broker they agree upon a version of the protocol to use. As a result, new features that rely on an updates to the protocol are only available when using both newer clients and newer brokers.
```

michaeljmarshall (approved):
LGTM. Sorry again for the delay, and thanks for your contribution!

onobc:
> LGTM. Sorry again for the delay, and thanks for your contribution!

No worries on the delay @michaeljmarshall and thank you for the help!

PR #418, opened 2023-02-16
Easier to copy&paste command for Windows

kuskmen:
I know its something small but it really should be able for windows people to simply copy and paste this shell command without needing to do any extra effort like removing the backslashes and making it one line because powershell doesn't support this. If you disagree feel free to decline this.

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`


tisonkun (approved):
Thanks for your contribution @kuskmen! LGTM.

I'd appreciate it if you can update `docs/getting-started-docker.md` file so that we have this improvement for all following versions also.

tisonkun:
Thank you! Merging...

PR #423, opened 2023-02-17
Fix wrong C++ client download page

BewareMyPower:
### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

The current C++ client download page is wrong, see:

It's because the table uses the `vtag` field to generate the **Release** and **Link** columns. 

https://github.com/apache/pulsar-site/blob/679b203bf481c552dde9d7dd4a932e7309e81d52/src/components/downloads.tsx#L177-L181

However, #419 set them as `3.1.x`. This PR uses the sub string of the `tagName` field instead of the `vtag` field, after the change the download page looks like:


tisonkun (changes requested):
This data file is used by https://pulsar.apache.org/release-notes/client-cpp/ also. I suggest you use `tagName` in the download page instead of change the vtag values.

BewareMyPower:
@tisonkun Fixed

tisonkun (approved):

PR #406, opened 2023-02-13
[fix][doc] fix broken links

momo-jun:
Fixes https://github.com/apache/pulsar/issues/19491 and all the broken links that have been warned during the local build for 2.8.x and later versions.

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

PR #435, opened 2023-02-24
[fix][doc] Fix broken links in legacy docs

momo-jun:
### Motivation
Following https://github.com/apache/pulsar-site/pull/406, this PR aims to clean up the existing 404 issues in Pulsar legacy docs before 2.8.0. 

### Modifications

The major reasons for broken links in the legacy doc sets and related modifications in this PR are:
* the linked file is missing for unknown reasons - the files are added back using the nearest doc version.
* the linked file is invalid because the feature is unsupported in those earlier versions - these links are removed.
I think some files have been overwritten by later versions unexpectly, for example, the doc version 2.2.0, 2.3.0 and 2.3.2 include content of TableView, which was introduced in 2.10.0.
* the linked file is in invalid format - these links are fixed.

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`


versioned_docs/version-2.2.0/helm-install.md
```diff
@@ -0,0 +1,40 @@
+---
+id: helm-install
```
Anonymitaet, line 2: Some features were added to earlier versions. Have you double checked this with engineers?

Anonymitaet (approved):

urfreespace (approved):

versioned_docs/version-2.2.0/helm-install.md
momo-jun: As mentioned in the PR description, I do think the doc sets for 2.2.0, 2.3.0 and 2.3.2 (and maybe more) have been overwritten by later versions unexpectedly during the website upgrade in last May and include many newly added features, e.g., TableView (introduced in 2.10.0) - I've removed the TableView related content though, but I believe there are much more, and challenging to check and clean one by one this way.

It would be more effective if there is a way to directly trace back the specific original doc versions; otherwise, those three doc sets have a huge gap in aligning with features. It's another thread, though. I discussed this with @urfreespace but the answer was negative.

So this PR aims to clean up 404 issues only. Despite this, I've checked with engineers about the availability of several features, e.g., schema, pulsar manager, security, etc.

PR #436, opened 2023-02-25
Updated the Hashing Scheme url

nikam14:
The URL was not working so updated it

### Documentation

- [ ] `doc` 
- [x] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/docs/concepts-messaging.md
+++ b/docs/concepts-messaging.md
@@ -904,7 +904,7 @@ If there is a key attached to message, the messages will be routed to correspond
 
 ### Hashing scheme
 
-{@inject: javadoc:HashingScheme:/client/org/apache/pulsar/client/api/HashingScheme} is an enum that represents sets of standard hashing functions available when choosing the partition to use for a particular message.
+{@inject: javadoc:HashingScheme:/client/2.4.2/org/apache/pulsar/client/api/HashingScheme} is an enum that represents sets of standard hashing functions available when choosing the partition to use for a particular message.
 
 There are 2 types of standard hashing functions available: `JavaStringHash` and `Murmur3_32Hash`.
 The default hashing function for producers is `JavaStringHash`.
--- a/docs/concepts-messaging.md
+++ b/docs/concepts-messaging.md
@@ -904,7 +904,7 @@ If there is a key attached to message, the messages will be routed to correspond
 
 ### Hashing scheme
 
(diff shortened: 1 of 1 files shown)


tisonkun (changes requested):
We shouldn't patch the link in this way. Instead, you can update the inject pattern to add the latest version when the version path is omitted.

The correct link for 2.11.x is https://pulsar.apache.org/api/client/2.11.x/org/apache/pulsar/client/api/HashingScheme.

IIRC `script/replace.js` has a similar logic.

docs/concepts-messaging.md
```diff
@@ -904,7 +904,7 @@ If there is a key attached to message, the messages will be routed to correspond
 
 ### Hashing scheme
 
-{@inject: javadoc:HashingScheme:/client/org/apache/pulsar/client/api/HashingScheme} is an enum that represents sets of standard hashing functions available when choosing the partition to use for a particular message.
+{@inject: javadoc:HashingScheme:/client/2.4.2/org/apache/pulsar/client/api/HashingScheme} is an enum that represents sets of standard hashing functions available when choosing the partition to use for a particular message.
```
tisonkun, line 907: Perhaps you can use:

```suggestion
[HashingScheme](/api/client/org/apache/pulsar/client/api/HashingScheme) is an enum that represents sets of standard hashing functions available when choosing the partition to use for a particular message.
```

And it will be resolved properly with `script/replace.js`

nikam14: Thank You for your guidance , I will improve it .

docs/concepts-messaging.md
```diff
@@ -904,7 +904,7 @@ If there is a key attached to message, the messages will be routed to correspond
 
 ### Hashing scheme
 
-{@inject: javadoc:HashingScheme:/client/org/apache/pulsar/client/api/HashingScheme} is an enum that represents sets of standard hashing functions available when choosing the partition to use for a particular message.
+[HashingScheme](https://pulsar.apache.org/api/client/2.11.x/org/apache/pulsar/client/api/HashingScheme) is an enum that represents sets of standard hashing functions available when choosing the partition to use for a particular message.
```
tisonkun, line 907:
```suggestion
[HashingScheme](/api/client/org/apache/pulsar/client/api/HashingScheme) is an enum that represents sets of standard hashing functions available when choosing the partition to use for a particular message.
```

I think the pattern is as this. I may try to migrate the post replace logic to a preview time avialable TSX component.

tisonkun (approved):

PR #439, opened 2023-02-28
fix rest api page crash

urfreespace:
close: https://github.com/apache/pulsar/issues/19644

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -54,10 +54,14 @@ export function setVersion(version) {
 }
 
 export function getVersion() {
-  if (/version=(\d+\.?\x?)+/.test(location.href)) {
-    return location.href.match(/version=(\d+\.?\x?)+/)[0];
+  try {
+    if (/version=(\d+\.?\x?)+/.test(location.href)) {
+      return location.href.match(/version=((\d+\.?\x?)+)/)[1];
+    }
+  } catch (error) {
+    console.error(error)
+    return "master";
   }
-  return "master";
 }
 
 export function getApiVersion(anchor) {
--- a/src/components/RestApi.js
+++ b/src/components/RestApi.js
@@ -1,4 +1,4 @@
-const React = require("react");
+import * as React from "react";
 import Layout from "@theme/Layout";
 import { getVersion, getApiVersion } from "../utils/index.js";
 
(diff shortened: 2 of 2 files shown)


src/utils/index.js
```diff
@@ -54,10 +54,14 @@ export function setVersion(version) {
 }
 
 export function getVersion() {
-  if (/version=(\d+\.?\x?)+/.test(location.href)) {
-    return location.href.match(/version=(\d+\.?\x?)+/)[0];
+  try {
+    if (/version=(\d+\.?\x?)+/.test(location.href)) {
+      return location.href.match(/version=((\d+\.?\x?)+)/)[1];
+    }
```
tisonkun, line 60: Is it possible we find the query part `version=...&apiversion=..` and match the version key-value pair exactly? Instead of using regexp which is brittle.

And, could you explain what is the inputs and outputs of this function?

urfreespace: this function is used to get version information from the URL `location.href`, for example: `https://pulsar.apache.org/admin-rest-api/?version=2.11.0&apiversion=v2#operation/getList`

urfreespace: > Is it possible we find the query part version=...&apiversion=.. and match the version key-value pair exactly? Instead of using regexp which is brittle.

NO, need to parse the query string yourself

tisonkun: OK. I may suggest to use `version=([0-9.x]+)` for comprehension. You can test it locally.

But +1 to merge as is.

tisonkun (approved):
See https://github.com/apache/pulsar-site/pull/439#discussion_r1119752768.

src/utils/index.js
urfreespace: got it, thanks.

urfreespace: https://github.com/apache/pulsar-site/pull/442 @tisonkun PTAL, thanks.

src/utils/index.js
```diff
@@ -54,10 +54,14 @@ export function setVersion(version) {
...
+  } catch (error) {
+    console.error(error)
+    return "master";
```
tisonkun, line 63: @urfreespace It seems now if we receive `version=master`, we return `undefined` instead of `master`?

This seems to be the cause of https://github.com/apache/pulsar/issues/19681.

cc @dave2wave

PR #441, opened 2023-02-28
Fix docker compose example to work with 0 errors

asafm:
# Changes
The current docker-compose example has several drawbacks:
- It doesn't use a fixed name; thus, when BK restarts, the cookie file stored in the data directory prevents BK from starting since it receives a different IP address hence a different name. Using "advertised" property prevents that.
- Memory is not set hence each container takes up 2 GB heap and 1gb direct memory - almost 9 GB for a single docker-compose file, making it fail without understanding why.

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`


asafm:
@BewareMyPower @hangc0276 Can you please double-check me?

tisonkun (approved):
LGTM.

It seems an improvement than fix? I'm thinking of whether we should fix also 2.11.x and early versions.

tisonkun:
Merging...

Thanks for your help!

PR #452, opened 2023-03-03
fix client cpp or python links

tisonkun:
* This closes https://github.com/apache/pulsar/issues/19675.
* This closes https://github.com/apache/pulsar/issues/19121.

Verified locally.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/scripts/replace.js
+++ b/scripts/replace.js
@@ -1,25 +1,12 @@
 const replace = require("replace-in-file");
 
-const fs = require("fs");
-
+const semver = require("semver");
 const CWD = process.cwd();
 const siteConfig = require(`${CWD}/docusaurus.config.js`);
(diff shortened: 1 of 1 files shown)


urfreespace (approved):

tisonkun:
The last commit touches a wide range of lines because we previously use `pulsar:version` in pulsar python client version while it's now not the case. And touching those files cause whitespace changes. But I think whitespace changes can be converged gradually.

For following reviewers, you can review this patch by commits instead of a full change set at once.

tisonkun:
CI passed and verified loaclly.

Merging...

PR #456, opened 2023-03-06
feat: reference page jump with version

urfreespace:
issue: https://github.com/apache/pulsar/issues/19676

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/scripts/replace.js
+++ b/scripts/replace.js
@@ -164,6 +164,7 @@ const latestVersion = getRealVersion(latestMajorRelease);
 const from = [
   /@pulsar:version_number@/g,
   /@pulsar:version@/g,
+  /@pulsar:version_origin@/g,
   /pulsar:binary_release_url/g,
   /pulsar:connector_release_url/g,
   /pulsar:offloader_release_url/g,
@@ -198,6 +199,7 @@ const options = {
   files: [`${nextDocsDir}/*.md`, `${nextDocsDir}/**/*.md`],
   from: from,
   to: [
+    `${latestVersion}`,
     `${latestVersion}`,
     `${latestVersion}`,
     binaryReleaseUrl(`${latestVersion}`),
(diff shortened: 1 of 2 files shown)


scripts/replace.js
```diff
@@ -164,6 +164,7 @@ const latestVersion = getRealVersion(latestMajorRelease);
 const from = [
   /@pulsar:version_number@/g,
   /@pulsar:version@/g,
+  /@pulsar:version_origin@/g,
```
tisonkun, line 167: I'm afraid that such variable cannot resolve 2.6~2.7 versioned docs to 2.6.x and 2.7.x properly. But if we change only for new versions, it can be reasonable.

urfreespace: yes, @momo-jun said only update the versions begin with `2.8.x`

momo-jun: @tisonkun It's acceptable, IMO, because we no longer actively improve the docs of those versions.
I will update the occurrences of links to the reference site in 2.8.x docs and later versions as soon as this PR can be merged.

tisonkun: Make sense. Although, this patch update for NEXT version only. @urfreespace would you update for 2.8.x, 2.9.x, 2.10.x and 2.11.x?

urfreespace: yeah, you are right, I just keep the script update and removed the update for next version, and @momo-jun will make another PR to update the docs, PTAL

momo-jun (approved):

tisonkun (changes requested):
It seems we don't touch any effective docs content now.

urfreespace:
> It seems we don't touch any effective docs content now.

@momo-jun will submit other PR to update the docs

urfreespace:
@tisonkun @momo-jun I commit a PR to update reference link for version 2.11.x https://github.com/apache/pulsar-site/pull/458 @momo-jun you can update other version docs of the rest

PR #477, opened 2023-03-21
Replace commerical page with tech docs

tisonkun:
### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`


tisonkun:
I'm going to prepare a new commit to trim all trailing space for docs. So that we don't review such large portion of whitespace changes later.

PR #481, opened 2023-03-21
[feat] Add canonical link for SEO

tisonkun:
This closes https://github.com/apache/pulsar/issues/18190.

Here are some screenshots for the result:

image
image

The "canonical" link are the only element in head metadata, no duplicate or confusion.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/src/theme/DocPage/index.tsx
+++ b/src/theme/DocPage/index.tsx
@@ -18,6 +18,8 @@ function createCanonicalHref(props: Props): string {
   const {siteConfig} = useDocusaurusContext();
   const {versionMetadata, location, match} = props;
   if (versionMetadata.version === 'current') {
+    // 1. The NEXT version docs should not forward to the latest stable version
+    // 2. Other plugins-doc instance (contribute, release-note, ...) happens to keep the link
     return siteConfig.url + useBaseUrl(location.pathname);
   }
   const basename = location.pathname.replace(match.path, '');
--- a/src/theme/DocPage/index.tsx
+++ b/src/theme/DocPage/index.tsx
@@ -5,7 +5,7 @@ import {
   DocsSidebarProvider,
   DocsVersionProvider,
   docVersionSearchTag,
-  useDocRouteMetadata, useDocsVersion,
+  useDocRouteMetadata,
 } from '@docusaurus/theme-common/internal';
 import DocPageLayout from '@theme/DocPage/Layout';
 import NotFound from '@theme/NotFound';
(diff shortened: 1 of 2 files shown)


src/theme/DocPage/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+        tag={docVersionSearchTag(
+          versionMetadata.pluginId,
+          versionMetadata.version,
+        )}
+      />
+      <PageMetadata>
+        {versionMetadata.noIndex && (
+          <meta name="robots" content="noindex, nofollow" />
+        )}
+        <link rel="canonical" href={createCanonicalHref(props)}/>
```
tisonkun, line 43: Only this line and the function `createCanonicalHref` are changed. Other lines are ejected internal logics which are already existing.

src/theme/DocPage/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+    <>
+      <SearchMetadata
+        version={versionMetadata.version}
+        tag={docVersionSearchTag(
+          versionMetadata.pluginId,
+          versionMetadata.version,
+        )}
+      />
+      <PageMetadata>
+        {versionMetadata.noIndex && (
```
tisonkun, line 40: Actually, if we properly configure `noIndex`, we can exclude old versions docs being indexed by the search engine.

The canonical link tweak can bring old links don't exist in latest version, but as we set up redirections properly, it's not a big problem.

cc @vinayrajsn @momo-jun @dave2wave thoughts?

urfreespace (approved):
awesome job, LGTM

tisonkun:
Merging...

Let's see if it helps.

src/theme/DocPage/index.tsx
tisonkun: Search engine reindex pages periodically.

momo-jun:
Hi @tisonkun, I wanted to ask a follow-up question on the logic that decides which canonical URLs to return in the Google results.

IIUC, the current adopted strategy is to enable two types of canonical links:
- `next` version uses `/next/` in the canonical URLs
- all the other versions use the latest stable version number in the canonical URLs

Then what logic decides the canonical URL to return in the Google results if the keyword exists in both the `next` version and earlier versions? In my observation, the behavior is random, so I'm wondering about the logic behind this.

For example, for "pulsar throttling", the next version shows: 
image

And for "pulsar functions", the latest stable version shows:
image

tisonkun:
> the behavior is random

Yes. IIRC there are some approaches we can hint the search engine about page "weight", but I'm not quite clear how we can implement it for our website now :)

PR #486, opened 2023-03-22
[improve][doc] Add schema support for python functions

momo-jun:
### Modification

Add docs for https://github.com/apache/pulsar/pull/18432
1. Add a table to show the supportability of schema types for language-specific function instances
2. Add a link to newly supported Python code examples.

Preview screenshot:

image

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`


momo-jun:
Hi @jiangpengcheng, @freeznet , can you pls review the doc PR?

jiangpengcheng (approved):

freeznet (approved):

docs/functions-develop-schema-registry.md
```diff
@@ -4,6 +4,19 @@ title: Use schema registry
...
 
-Pulsar has a built-in schema registry and is bundled with popular schema types, such as Avro, JSON and Protobuf. Pulsar Functions can leverage the existing schema information from input topics and derive the input type. The schema registry applies to output topics as well.
+Pulsar has a built-in [schema registry](schema-overview.md) and is bundled with popular [schema types](schema-understand.md#schema-type). Pulsar Functions can leverage the existing schema information from input topics and derive the input type. The schema registry applies to output topics as well.
 
-For more details, refer to [code example](https://github.com/apache/pulsar/blob/master/pulsar-functions/java-examples/src/main/java/org/apache/pulsar/functions/api/examples/AutoSchemaFunction.java).
+The following table outlines the supportability of schema types in Pulsar Functions.
+
+| Schema Type    | Java Function | Python Function | Go Function |
+|----------------|---------------|-----------------|-------------|
+| String         | ✅             | ✅               |             |
```
Anonymitaet, line 13:
```suggestion
| String         | ✅             | ✅               | Not supported            |
```

Suggest indicates "not supported" (or use ❌ to keep consistent) explicitly to reduce users' guesswork. 

(In this case, "Not supported" or "❌" is not a negative signal, it just shows the status and fact)

momo-jun: I understand your concern, but I don't think it will make readers guess without those inverted indicators; on the contrary, IMO, adding "not supported" is a bit redundant. What's worse, to some users, ❌ also means "selected" which leads to confusion.

Googled a few more references from MS and Splunk:
image
image

Anonymitaet: If there is no ✅, users (like German) might take ❌ as "supported". Since there is ✅, why do users still take ❌ as "supported"?

momo-jun: I got the feedback from a customer-facing team when both indicators were placed in a table years ago. They felt confused because they weighed X higher than any others to represent YES in a table.
So, taking this into considerations, I don't see the need for adding inverted ones here. It's clear and specific enough for users to read the table.

Anonymitaet: From my past experience, some users asked me what the meaning of "empty"? They take "empty" as "maintainers **forget** to fill info" and "**unsure** whether it means unsupported or not" because **explicit** answers are not given — this is the [guesswork](https://github.com/apache/pulsar-site/pull/486#discussion_r1145530721) that happened among users and should be avoided from a rigorous perspective.

momo-jun: Thanks for sharing more ideas! 
I can understand the guesswork of an empty cell that you mentioned when it comes to a parameter table in terms of name, description, default value, etc. The case in this PR is different, which is all about a Yes/No answer, without other options. 

Regarding this example https://github.com/apache/pulsar-site/pull/486#discussion_r1145775435, I think it's ambiguous. I guess the meaning of the hollow circle sign is "partially supported" but I'm not sure.

I will merge the PR and revisit this thread if I find a compromised way to balance different user experiences :)

PR #472, opened 2023-03-16
[improve][doc] Add configurations doc for the python client

RobertIndie:
Here are specific links for these configuration doc:
* Client: https://pulsar.apache.org/api/python/3.1.x/pulsar.Client.html#__init__
* Producer: https://pulsar.apache.org/api/python/3.1.x/pulsar.Client.html#create_producer
* Consumer: https://pulsar.apache.org/api/python/3.1.x/pulsar.Client.html#subscribe
* Reader: https://pulsar.apache.org/api/python/3.1.x/pulsar.Client.html#create_reader 

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

PR #465, opened 2023-03-13
[improve][doc] Add configurations doc for the cpp client

RobertIndie:
### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

PR #393, opened 2023-01-30
[refactor][doc] Refactor the information architecture of client library docs

momo-jun:
### Motivation

The pull request is part of [PIP-228](https://github.com/apache/pulsar/issues/18822) - Implement the IA changes in the client library docs to map content to the user journey and learning path.

### Modifications

1. Create "Get started" topics for language-specific client libraries to onboard users with a clear map and explicit information.
2. Organize advanced information by user tasks, and show language-specific code snippets using the tab style. 

The IA and related pages have been changed significantly in terms of structures and headings. You can check out the branch to your local for a thorough preview. This is only one snapshot.

Note that this PR only focuses on the structure improvements, while the detailed content itself definitely needs to be reviewed and enriched through a series of pull requests later.

image

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

PR #487, opened 2023-03-22
[improve][doc] Relocate client config links and refresh client API doc links

momo-jun:
This PR is a follow-up to the Pulsar client doc refactoring implemented in https://github.com/apache/pulsar-site/pull/393.

Thank @RobertIndie for adding the missing configs for [C++](https://github.com/apache/pulsar-site/pull/465/)/[Python](https://github.com/apache/pulsar-site/pull/472)/[Go](https://github.com/apache/pulsar-site/pull/479) clients.

Since the reference config docs (Java/C++/Python/Go) are auto-generated from code and shown as links in the docs, we can relocate these ref links to make the structure more flattened and straightforward. 

The [existing static configs for Node.js client](https://pulsar.apache.org/docs/next/client-libraries-node-configs/) are retained in the docs, since we don't have auto-generated API docs for [Node.js](https://github.com/apache/pulsar/issues/19876) (and [C#](https://github.com/apache/pulsar/issues/19877)) in place at this moment, but it can be removed as soon as the gap is fulfilled.

### Modification

1. Relocate the links of configs of Java/C++/Python/Go clients.
2. Relocate the Java client memory config info to the Reference site.
3. Refresh the links pointing to client API docs.
4. Relocate the API overview topic to the Reference category and restructure it with the client API overview added.
5. Minor fixes.

Preview screenshot of the Reference site:
image

Preview screenshot of the relocated and updated API overview topic:
image

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`


static/reference/next/client/README.md
```diff
@@ -1 +1,28 @@
...
+- [Client configurations](/next/client/client-configuration-client)
+- [Producer configurations](/next/client/client-configuration-producer)
+- [Consumer configurations](/next/client/client-configuration-consumer)
+- [Reader configurations](/next/client/client-configuration-reader)
+
+### Set client memory allocator configs using Java system properties
+
+You can set the client memory allocator configurations through Java properties.<br />
+
+| Property | Type | <div style="width: 110pt"> Description </div> | Default | Available values |
```
momo-jun, line 14: `110pt` is a compromised value when I was trying to expand the width of this column to include more characters in one row. Higher values will expand the entire table, and users have to scroll horizontally to see all the information. If there are any other better solutions, please let me know.

docs/client-libraries-java.md
```diff
@@ -29,12 +29,12 @@ The following table outlines the API packages and reference docs for Pulsar Java
 
 Package | Description | Maven Artifact
 :-------|:------------|:--------------
-[`org.apache.pulsar.client.api`](/api/client) | [Java client API](/api/client/) | [org.apache.pulsar:pulsar-client:@pulsar:version@](http://search.maven.org/#artifactdetails%7Corg.apache.pulsar%7Cpulsar-client%7C@pulsar:version@%7Cjar)
+[`org.apache.pulsar.client.api`](/api/client) | [Java client API](client-api-overview.md) | [org.apache.pulsar:pulsar-client:@pulsar:version@](http://search.maven.org/#artifactdetails%7Corg.apache.pulsar%7Cpulsar-client%7C@pulsar:version@%7Cjar)
```
tisonkun, line 32: I don't think `client-api-overview` is the correct page to be linked here. Actually, I'd suggest to merge that page into clent-libraries-overview.

static/reference/next/client/README.md
```diff
@@ -1 +1,28 @@
-> Docs for Pulsar Java Client Configuration.
+## Client Configurations
+
+If you create a Java client, producer, consumer, or reader, you can use the `loadConf` configurations.
+
+- [Client configurations](/next/client/client-configuration-client)
```
momo-jun, line 5: @urfreespace  I'm wondering whether we can use `@pulsar:version_origin@` to replace `/next`? This cannot be tested out in my local build.

docs/client-libraries-java.md
momo-jun: I agree it doesn't make 100% sense. The initial motivation to change the link is because `/api/client` has been linked in the first column. I think it would be redundant to provide the reference link again.
@Anonymitaet WDYT?

Anonymitaet: Since now we only have 1-page content for https://pulsar.apache.org/docs/next/client-api-overview/, it can go to `Reference > API > Client` with API reference doc links for three languages (Java, Python, C++) since they are more likely to belong to "Reference" topic rather than "Concept" topic (Client Libraries > Overview).

momo-jun: To me, the paragraph in https://pulsar.apache.org/docs/next/client-api-overview/ looks more like a concept introduction, where we provide descriptions and an illustration for explanations.
I intend to relocate it to 'Pulsar API Overview' since it's the most related topic. @Anonymitaet - I wonder whether it conflicts with your plan on improving/expanding API docs. If yes, I think `Client library Overview` can be an alternative.

Anonymitaet: Feel free to make your desired changes.

static/reference/next/client/README.md
momo-jun: Using the new var `@pulsar:version_reference@` to replace `next`.

docs/pulsar-api-overview.md
```diff
@@ -1,22 +1,39 @@
...
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 ````
 
-Pulsar is a messaging and streaming platform that scales across organizations of all sizes. Pulsar APIs are the core building blocks of Pulsar, which allow you to interact with Pulsar or administer Pulsar clusters. Pulsar APIs consist of the following types:
+Pulsar is a messaging and streaming platform that scales across organizations of all sizes. 
 
-- [Pulsar client APIs](client-api-overview.md)
-
-- [Pulsar admin APIs](admin-api-overview.md)
```
Anonymitaet, line 16: Why delete the following contents? They are technically correct and it explains why "Pulsar client API" and "Pulsar admin API" are hosted inside this topic.

```
Pulsar APIs consist of the following types:

- [Pulsar client APIs](client-api-overview.md)

- [Pulsar admin APIs](admin-api-overview.md)
```

momo-jun: It's a compromise. 
1. The links are too close (to the following two targets) to redirect, especially for the admin API section (newly added heading2), which provides another link to the overview topic. It's just not a good idea to let users click twice for a single piece of information. So I select to keep the link under the heading2.
2. Both the illustration and the following headings can help them get to this point.

momo-jun: Thanks for your suggestions added in the screenshot. I can move the API overview to the top, but I prefer not to add a standalone heading4 for REST APIs inside the topic. Similar to the bullet style, a single one doesn't make too much sense but has the chance to confuse users. We can revisit this if we have at least two items in parallel.

Anonymitaet: > a single one doesn't make too much sense but has the chance to confuse users.

The parallel structure (Pulsar APIs and REST APIs) confuses users as well.

If not embedding it to the "Pulsar APIs" topic, how about optimizing the left TOC as below?

```
# Reference

## Pulsar APIs
### REST APIs

## Terminology
## Pulsar CLI tools
## ...
```

momo-jun: > The parallel structure (Pulsar APIs and REST APIs) confuses users as well.

Yes, I agree. I'll change the sidebar label as a simple way to reconcile.

sidebars.json
```diff
@@ -517,11 +492,12 @@
         "id": "reference-landing"
       },
       "items": [
+        "pulsar-api-overview",
+        "reference-rest-api-overview",
```
Anonymitaet, line 496: Since `pulsar-api-overview` is in 1st level, then`reference-rest-api-overview` should be in 2nd level, no?

Anonymitaet (approved):

tisonkun (approved):

PR #493, opened 2023-03-28
[improve][doc] Optimize Admin API > Features > Brokers

Anonymitaet:
## What issues does this PR solve?

There are some issues in the current [Admin API > Features > Brokers](https://pulsar.apache.org/docs/next/admin-api-brokers/) doc. I’ve summarized them with the corresponding solutions below.

Issues | Solutions
-- | --
Some docs are inaccurate and make users confused. | Correct the description.
Some docs are redundant and scattered. | Consolidate and optimize the explanations.
Some docs are wrongly placed. <br><br>For example, "Concept"-type docs are shown in Sub-features, while they should be located elsewhere in Pulsar docs. | Move and consolidate them to the correct place. <br><br>For example, move "Concept"-type docs to [Reference > Terminology](https://pulsar.apache.org/docs/next/reference-terminology/).
Some operations are in chaotic order. | Categorize operations logically to match users' learning pace. <br><br>For example, classify operations based on the CRUD sequence.

### ✅ High-level structure changes

- 🍊Old structure
image

- 🍊New structure
image

=================

- [x] `doc` 

Ping @momo-jun @D-2-Ed to review, thank you!


Anonymitaet:
Previews look good:

image

image

Anonymitaet:
@BewareMyPower Could you please review this PR from a technical perspective? Thank you!

docs/admin-api-brokers.md
```diff
@@ -87,39 +69,53 @@ Fetch the information of the leader broker, for example, the service url.
...
 </Tabs>
 ````
 
-#### list of namespaces owned by a given broker
+## Update broker conf 
+
+You can update broker configurations using one of the following ways:
 
-It finds all namespaces which are owned and served by a given broker.
+- Supply [configurations](https://pulsar.apache.org/reference/#/@pulsar:version_origin@/config/reference-configuration-broker) when starting up brokers.
```
momo-jun, line 110:
```suggestion
- Supply [configurations](pathname:///reference/#/@pulsar:version_origin@/config/reference-configuration-broker) when starting up brokers.
```

docs/admin-api-brokers.md
```diff
@@ -87,39 +69,53 @@ Fetch the information of the leader broker, for example, the service url.
...
+
+- [Update configurations dynamically](#update-broker-conf-dynamically) **when running brokers**.
+
+  Since all broker configurations in Pulsar are stored in ZooKeeper, configuration values can also be dynamically updated when brokers are running. When you update broker configurations dynamically, ZooKeeper will notify the broker of the change and then the broker will override any existing configuration values.
+
+### List updatable broker conf
```
momo-jun, line 116: It seems more like an optional step before users dynamically update broker conf. The same applies to `List updated broker conf`. Logically, they are talking about the `list` feature, not `update`, but can be leveraged before/after updating broker conf.

So I suggest a tweak on heading 2 and heading 3, using either of the following. 

**OptionA:**
```
## List updatable broker conf
...
## List updated broker conf
...
## Update broker conf dynamically
... // link to the the List headings
```
**OptionB:**
```
## Update broker conf dynamically
1. List updatable broker conf
2. ...
3. List updated broker conf
```

The difference between A and B is whether we want to highlight the feature of `list`. WDYT?

docs/admin-api-brokers.md
momo-jun: If we tweak the heading by adding `dynamically`, then this one can become a supplemented tip.

Anonymitaet: Since the heading is "Update broker conf", so two ways should be treated equally from a doc structure perspective, it does not deserve a tip.

momo-jun: This suggestion is given based on the [proposed structure tweak](https://github.com/apache/pulsar-site/pull/493#discussion_r1150112037). 

`Admin-api-brokers` is a task topic for using APIs to implement features around brokers. The link to the reference site supplements the implementation and is somehow different from the primary story. So my idea is we can make it a tip  to provide the other static way when we use `Update broker conf dynamically` as heading 2.

momo-jun (approved):
Generally looks good.
Structure-wise, I've left my suggestions.

BewareMyPower (approved):

docs/admin-api-brokers.md
Anonymitaet: I'd prefer not to use an ordered list here. Reasons:

1. This page is a "reference" topic rather than a "task" topic.
  
    In a "task" topic, ordered lists in headings make sense because the sequence matters.

image
    
   But in a "reference" topic, it's rare to use ordered lists for operations because sequence does not matter most. 

2. `List updatable broker conf`, `Update broker conf dynamically`, and `List updated broker conf` are not complex operations. Users can know their logic at 1st glance.

momo-jun: It's not a big issue, but I wanted to make myself clear for your reference :)
The possible confusion exists because the relationship between heading 2 and nested three heading 3 is not self-explained, making readers to think and guess (I did)... e.g.,
* Are those heading3 separate sub-features? Why 'list' appears here?
* If a `dynamic` item is a sub-feature, why a contrary one is absent? It's not MECE.
* What's the relationship amount the three? Do they all serve the heading2 `Update broker conf`? After reading the content, I found the `list` ones serve `Update broker conf dynamically`.

image

That's why I suggest changing the heading3 into plain-text ordered steps. BTW, I don't see this category as a reference since we have a doc set including overview and get-started topics for it. It's more like a key feature. The topic title "Brokers" means "Manage brokers" in my eyes.

Anonymitaet: > Are those heading3 separate sub-features? Why 'list' appears here?

Yes, they are all sub-features.

> If a dynamic item is a sub-feature, why a contrary one is absent? It's not MECE.

Because we do not have the contrary feature.

> What's the relationship amount the three? Do they all serve the heading2 Update broker conf? After reading the content, I found the list ones serve Update broker conf dynamically.

They all work for heading2 `update broker conf` because e only have `update broker dynamically` and do not have the contrary one. So`update broker conf` is an appropriate heading2 for all 3 operations.

> I don't see this category as a reference since we have a doc set including overview and get-started topics for it.

Don't be puzzled by the title, location, and doc structure. It looks like a "concept" or "get started" topic at 1st glance. 

But if taking a closer and careful look, you'll find that the docs are talking about "command references" rather than "what is it" or "how to use it". So it's a "reference" topic in essence.

PR #506, opened 2023-04-06
[feat] noIndex unmaintained pages for SEO

tisonkun:
I still searched some ancient versions on Google. Perhaps the canonical link trick doesn't work or it needs more time to propagate. But let's add more meta to help the search engine not index unmaintained pages.

2.9.x (unmaintained):

image

2.10.x, 2.11.x, and NEXT don't have this meta tag.

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/src/theme/DocPage/index.tsx
+++ b/src/theme/DocPage/index.tsx
@@ -13,6 +13,7 @@ import SearchMetadata from '@theme/SearchMetadata';
 import type {Props} from '@theme/DocPage';
 import useDocusaurusContext from "@docusaurus/useDocusaurusContext";
 import useBaseUrl from "@docusaurus/useBaseUrl";
+import semver from "semver/preload";
 
 function createCanonicalHref(props: Props): string {
   const {siteConfig} = useDocusaurusContext();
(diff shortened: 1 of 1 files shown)


momo-jun:
Is this change to not index docs for versions earlier than the latest stable one? The PR title says 'unmaintained' which means the versions earlier than 2.8.x.

tisonkun:
Unmaintained is defined in https://pulsar.apache.org/contribute/release-policy/. Currently, it's 2.9.x and before.

Let's hard code for now and if it works, we can refactor the logic to auto adjust when new releases happen.

momo-jun (approved):

tisonkun:
Merging and rebuilding...

For people who are using an unmaintained version, he/she can still search within the website. We just don't recommend or propagate it to the search engine. After we release the first LTS version, the case that many users use an unmaintained version should ease. Also, most doc pages are still existing on maintained pages, so a specific topic is still searchable.

PR #507, opened 2023-04-07
revise ambiguous statements

rupertw:
Originally the statement may means shared subscriptions include key-shared subscriptions, but shared and key-shared are totally different subscription types.

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`


tisonkun (approved):
Thank you! LGTM.

PR #513, opened 2023-04-12
[feat] Add client matrix page

urfreespace:
This closes https://github.com/apache/pulsar/issues/17098

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Screenshot
matrix


momo-jun:
Thanks Lili. Add a few questions:
1. How to maintain the status in these tables? We need a user-friendly way so that maintainers can update the table on a regular basis.
2. Can we fold each table? Or add hot links on the right side so that users can read the whole page at a glance.
3. Can we add a scroll bar horizontally and vertically? Now the consumer matrix cannot be displayed in one screen, so we may lose the table header when looking at the bottom row.

BTW, We plan to add reader/tableview in the next delivery phase and can add them when iterating the matrix.

urfreespace:
> Thanks Lili. Add a few questions:
> 
> 1. How to maintain the status in these tables? We need a user-friendly way so that maintainers can update the table on a regular basis.
> 2. Can we fold each table? Or add hot links on the right side so that users can read the whole page at a glance.
> 3. Can we add a scroll bar horizontally and vertically? Now the consumer matrix cannot be displayed in one screen, so we may lose the table header when looking at the bottom row.
> 
> BTW, We plan to add reader/tableview in the next delivery phase and can add them when iterating the matrix.

done, PTAL, thanks @momo-jun

urfreespace:
image

Hi, @tisonkun would you pls help take a look at this CI problem? thanks very much

tisonkun:
@urfreespace fixed. It's because we now check-in yarn.lock to leverage yarn 3 pkg manager.

urfreespace:
> @urfreespace fixed. It's because we now check-in yarn.lock to leverage yarn 3 pkg manager.

thanks

docusaurus.config.js
```diff
@@ -393,6 +393,16 @@ module.exports = {
                 sidebarPath: false,
             }),
         ],
+        [
+          "content-docs",
```
tisonkun, line 397: If it's a standalone page, we may not need a docs plugin instance.

client-feature-matrix/index.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+### Producer
+
+<Matrix data={producer} name="producer" />
+
+### Consumer
+
+<Matrix data={consumer} name="consumer" />
+
+See details and full version [here](https://docs.google.com/spreadsheets/d/1YHYTkIXR8-Ql103u-IMI18TXLlGStK8uJjDsOOA0T20/edit?usp=sharing)
```
tisonkun, line 29: After we land this page, shall this page be the source of truth so that we don't refer to the gdoc? Otherwise, why do we add such a page?

tisonkun (changes requested):
Thanks for your work! Comments inline. can be a blocker.

client-feature-matrix/index.mdx
momo-jun: Yes, it should be the source of truth. However, the page is a compact version at the pilot stage.
In case some users or maitainers still need more information from the working sheet, we attach the link here to not lose the connection with the full version.

tisonkun: Weird. Then we create two similar but different matrices, which seems not a good idea.

tisonkun: For example, I'm a developer, I cannot trust the page but have to look at the gdoc version because any updates will go there first. As a user, it's the same. This patch introduces a _maybe_ up-to-date version but any updates and nightly changes happen on the gdoc. Then I suggest we use only the gdoc and that should not worse the case.

momo-jun: Maybe I didn't make it clear. In this PR, we only add the matrix for Client, Producer, and Consumer. With more feedback/voice from the community, we will iterate and continue to add the matrix for Reader and TableView. After that, we can move the link.

tisonkun: Thanks for your explanation! I'd prefer to remove the link so that it's clear that we'd like to move to this page. Instead of leaving a workaround and it's highly possible the intermediate situation remains for a long time.

momo-jun (approved):
Looks good to me. Nice work! 
I will follow up with a few updates on the doc side.

client-feature-matrix/index.mdx
momo-jun, line 29:
```suggestion
See details and full version [here](https://docs.google.com/spreadsheets/d/1YHYTkIXR8-Ql103u-IMI18TXLlGStK8uJjDsOOA0T20/edit?usp=sharing).
```

docusaurus.config.js
urfreespace: it's a simplified way to implement the below effect (the absolute same TOC effect with the docs page)

image

@tisonkun

tisonkun: Thanks for your explanation! I agree with this trade-off now :D

PR #514, opened 2023-04-12
[improve][doc] Add explanations and illustrations for failover

Anonymitaet:
This PR adds docs  for https://github.com/apache/pulsar/pull/19502

- [x] `doc` 

Ping @D-2-Ed for editorial review, thanks!


Anonymitaet:
All previews look good:

image

image

image

Anonymitaet:
For reviewers: the images look a little large, but it seems that the "Docusaurus markdown" does not support image size customization, and the default image size is displayed. 

I've tried HTML syntax, but it fails to render since New Docusaurus does not support HTML (while it used to). Its official doc indicates that "use markdown rather than HTML as much as possible".

So let's use these images now. I'll resize them once there's a feasible solution.

Anonymitaet:
@codelipenghui thanks for your technical guidance! Could you please help review this PR?

docs/concepts-messaging.md
```diff
@@ -515,16 +515,64 @@ Exclusive is the default subscription type.
...
 
-* For partitioned topics, the broker will sort consumers by priority level and lexicographical order of consumer name. The broker will try to evenly assign partitions to consumers with the highest priority level.
-* For non-partitioned topics, the broker will pick consumers in the order they subscribe to the non-partitioned topics.
+A master consumer is picked for a non-partitioned topic or each partition of a partitioned topic and receives messages. 
 
-For example, a partitioned topic has 3 partitions, and 15 consumers. Each partition will have 1 active consumer and 4 stand-by consumers.
+When the master consumer disconnects, all (non-acknowledged and subsequent) messages are delivered to the next consumer in line.
 
-In the diagram below, **Consumer A** is the master consumer while **Consumer B** would be the next consumer in line to receive messages if **Consumer A** is disconnected.
+##### Failover | Partitioned topics
```
Anonymitaet, line 524: For reviewers: 

To improve legibility and readability, I add `|` in 2 headings because 4-level headings and 5-leave headings look the same size on the webpage (this is a Docusarus limitation).

docs/concepts-messaging.md
```diff
@@ -515,16 +515,64 @@ Exclusive is the default subscription type.
...
+  
+  Each partition has 1 active consumer and 1 stand-by consumer. 
+  
+    - For p0, Consumer A is the master consumer, while Consumer B would be the next consumer in line to receive messages if Consumer A is disconnected.
+
+    - For p1, Consumer C is the master consumer, while Consumer D would be the next consumer in line to receive messages if Consumer C is disconnected.
+
+  ![Failover subscriptions](/assets/pulsar-failover-subscriptions-4.svg)
+
+- If the number of a partitioned topic is **greater** than the number of consumers:
```
momo-jun, line 544:
```suggestion
- If the number of partitions in a partitioned topic is **greater** than the number of consumers:
```
The `equal` scenario is missed here and needs to be clarified.

docs/concepts-messaging.md
```diff
@@ -515,16 +515,64 @@ Exclusive is the default subscription type.
...
+##### Failover | Partitioned topics
 
-![Failover subscriptions](/assets/pulsar-failover-subscriptions.svg)
+For partitioned topics, the broker sorts consumers by priority level and lexicographical order of consumer name. 
+
+The broker tries to evenly assign partitions to consumers with the highest priority level. 
+
+A consumer is selected by running a module operation `mod (partition index, consumer index)`.
+
+- If the number of a partitioned topic is **less** than the number of consumers:
```
momo-jun, line 532:
```suggestion
- If the number of partitions in a partitioned topic is **less** than the number of consumers:
```
Is it what you wanted to convey?

static/assets/pulsar-failover-subscriptions-1.svg
momo-jun, line 1: The lines in the light yellow color are a bit too light to see (kind of blend in the background color).

docs/concepts-messaging.md
Anonymitaet: I do not add `equal` because users can figure it (how consumers are assigned) out quickly because it's easy and obvious. Let me double-check with @codelipenghui.

Anonymitaet: Hi @codelipenghui 

1️⃣ For `equal` scenario, one consumer will be assigned to one partition, right?
e.g., if there are 3 partitions and 3 consumers, then Consumer A is assigned to p0, Consumer B is assigned to p1, and Consumer C is assigned to p2. It's an easy one.

2️⃣ So for the `equal` case, should we add an example and illustration for it? (like what we do for `greater` and `less`) 
Does the user know the behavior as stated in 1️⃣? If so, it's redundant to add those since it's easy and obvious.

docs/concepts-messaging.md
Anonymitaet: Oops! Thanks! Updated.

codelipenghui (approved):

static/assets/pulsar-failover-subscriptions-1.svg
Anonymitaet: Agree. In fact, I've tried two kinds of yellow.

Light yellow: it's more aesthetic but hard to recognize. 
Dark yellow: it's ugly but easy to tell.

Previously I chose light yellow because I like beautiful stuff 😝
But just now I updated it to dark yellow to fit into readers' shoes and improve legibility since "usefulness" matters most in this case.

image

Anonymitaet:
@Huanli-Meng Could you please review this PR? Thank you!

docs/concepts-messaging.md
```diff
@@ -515,16 +515,64 @@ Exclusive is the default subscription type.
...
+A master consumer is picked for a non-partitioned topic or each partition of a partitioned topic and receives messages. 
 
-For example, a partitioned topic has 3 partitions, and 15 consumers. Each partition will have 1 active consumer and 4 stand-by consumers.
+When the master consumer disconnects, all (non-acknowledged and subsequent) messages are delivered to the next consumer in line.
 
-In the diagram below, **Consumer A** is the master consumer while **Consumer B** would be the next consumer in line to receive messages if **Consumer A** is disconnected.
+##### Failover | Partitioned topics
 
-![Failover subscriptions](/assets/pulsar-failover-subscriptions.svg)
+For partitioned topics, the broker sorts consumers by priority level and lexicographical order of consumer name.
```
Huanli-Meng, line 526:
```suggestion
For partitioned topics, the broker sorts consumers by priority and lexicographical order of consumer name. 
```

docs/concepts-messaging.md
```diff
@@ -515,16 +515,64 @@ Exclusive is the default subscription type.
...
-For example, a partitioned topic has 3 partitions, and 15 consumers. Each partition will have 1 active consumer and 4 stand-by consumers.
+When the master consumer disconnects, all (non-acknowledged and subsequent) messages are delivered to the next consumer in line.
 
-In the diagram below, **Consumer A** is the master consumer while **Consumer B** would be the next consumer in line to receive messages if **Consumer A** is disconnected.
+##### Failover | Partitioned topics
 
-![Failover subscriptions](/assets/pulsar-failover-subscriptions.svg)
+For partitioned topics, the broker sorts consumers by priority level and lexicographical order of consumer name. 
+
+The broker tries to evenly assign partitions to consumers with the highest priority level.
```
Huanli-Meng, line 528: I would suggest to use "priority" rather than priority level.

docs/concepts-messaging.md
momo-jun: Oh, I missed the following statement but paid attention to the two IF conditions. Maybe adding a descriptive example starting with `-` would be good enough to show all the three conditions. 

> The broker tries to evenly assign partitions to consumers with the highest priority level.

momo-jun (approved):

docs/concepts-messaging.md
```diff
@@ -515,16 +515,64 @@ Exclusive is the default subscription type.
...
+  
+    - For p0, Consumer A is the master consumer, while Consumer B would be the next consumer in line to receive messages if Consumer A is disconnected.
+
+    - For p1, Consumer C is the master consumer, while Consumer D would be the next consumer in line to receive messages if Consumer C is disconnected.
+
+  ![Failover subscriptions](/assets/pulsar-failover-subscriptions-4.svg)
+
+- If the number of partitions in a partitioned topic is **greater** than the number of consumers:
+  
+  For example, in the diagram below, this partitioned topic has 9 partitions and 3 consumers.
```
Huanli-Meng, line 546: Out of curiosity, in this case, each partition has one consumer, so how to realize failover?

docs/concepts-messaging.md
```diff
@@ -515,16 +515,64 @@ Exclusive is the default subscription type.
...
+
+    - For p1, Consumer C is the master consumer, while Consumer D would be the next consumer in line to receive messages if Consumer C is disconnected.
+
+  ![Failover subscriptions](/assets/pulsar-failover-subscriptions-4.svg)
+
+- If the number of partitions in a partitioned topic is **greater** than the number of consumers:
+  
+  For example, in the diagram below, this partitioned topic has 9 partitions and 3 consumers. 
+  
+  - p0, p3, and p6 are assigned to consumer A.
```
Huanli-Meng, line 548: consumer A or Consumer A, should keep consistent through the whole doc

docs/concepts-messaging.md
```diff
@@ -515,16 +515,64 @@ Exclusive is the default subscription type.
...
+  
+  Consumer A is the master consumer, while Consumer B would be the next consumer in line to receive messages if Consumer A is disconnected.
+
+  ![Failover subscriptions](/assets/pulsar-failover-subscriptions-2.svg)
+
+- If there are multiple non-partitioned topics, a consumer is selected based on **consumer name hash** and **topic name hash**. The client uses the same consumer name to subscribe to all the topics.
+
+  For example, in the diagram below, there are 4 non-partitioned topics and 2 consumers. 
+  
+  - The non-partitioned topic 1 and non-partitioned topic 4 are assigned to Consumer B.
```
Huanli-Meng, line 571: I think we can remove "non-partitioned" as you have said in the leading sentence.

Huanli-Meng (commented):
Some minor comments. PTAL

docs/concepts-messaging.md
Anonymitaet: Oops! I choose lowercase and keep them consistent. Updated.

docs/concepts-messaging.md
Anonymitaet: I prefer to keep it to make the info more clear (it's not a redundancy)

docs/concepts-messaging.md
Anonymitaet: Agree. Updated.

PR #526, opened 2023-04-19
Update the process to upload the GPG key

BewareMyPower:
See the discussion here https://lists.apache.org/thread/f9w430oqpm0g72b1htwbtc8y3mfqf8r6

In addition, download the `KEY` file before appending a KEY because `svn up` might not download the file.

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/contribute/create-gpg-keys.md
+++ b/contribute/create-gpg-keys.md
@@ -86,26 +86,7 @@ New **RSA** keys generated should be at least **4096** bits.
 
 ## Appending the key to KEYS files
 
-The GPG key needs to be appended to `KEYS` file that is stored in 2 SVN locations, one for proper releases and one for the release candidates.
-
-The credentials for SVN are the usual Apache account credentials.
-
-```shell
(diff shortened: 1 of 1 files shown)


contribute/create-gpg-keys.md
```diff
@@ -118,6 +99,7 @@ You should ask a PMC member to complete this step.
 svn co https://dist.apache.org/repos/dist/release/pulsar pulsar-dist-release-keys --depth empty
```
tisonkun, line 99:
```suggestion
svn co https://dist.apache.org/repos/dist/release/pulsar pulsar-dist-release-keys --depth files
```

I actually use files so the only file `KEYS` is downloaded properly. Maybe you can give it a try and see if it's a better/more concise approach.

BewareMyPower: Yes, it works for me.

tisonkun (approved):

PR #536, opened 2023-04-24
Update io-debezium-source.md

raymondBourges:
Without "topicName" in YAML definition file "bin/pulsar-admin source localrun --source-config-file debezium-postgres-source-config.yaml" exits with a "Topic name cannot be null" error.

Note : I just can give topic name online (without using a YAML file) within --sourceConfig parameter of bin/function-localrunner command but this command is not present in this documentation. Here, just bin/pulsar-admin is documented but it doesn't accept topic name as parameter.

### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`


Anonymitaet:
@liangyepianzhou Could you please review this PR from a technical perspective? TIA!

versioned_docs/version-2.11.x/io-debezium-source.md
```diff
@@ -268,6 +268,7 @@ You can use one of the following methods to create a configuration file.
   inputs: [ "debezium-postgres-topic" ]
   archive: "connectors/pulsar-io-debezium-postgres-@pulsar:version@.nar"
   parallelism: 1
+  topicName: "public/default/log-debezium-postgres"
```
liangyepianzhou, line 271: Do you verify it?
I see the configuration in Pulsar just "log-debezium-postgres" [here](https://github.com/apache/pulsar/blob/master/pulsar-io/debezium/postgres/src/main/resources/debezium-postgres-source-config.yaml#L35).

image

tisonkun: It seems the tenant and namespace part can be auto appended so both is correct?

tisonkun:
We actually have a similar PR for this topic - https://github.com/apache/pulsar-site/pull/388.

@raymondBourges

raymondBourges:
@tisonkun,
"to verify your change" --> I'm not sure what you expect from me.
I just wanted to contribute to the documentation where in https://pulsar.apache.org/docs/2.11.x/io-debezium-source/#example-of-postgresql the example that is given does not work since " topicName" is required.
This is to prevent other Pulsar users from wasting time understanding why the CDC doesn't work if you strictly follow the documentation.
Note that I only contributed on the postgres part because it is the only one I tested, but it is probably necessary to have a look to the examples for the other bases.
Also note that the example given by Xiangying is correct but does not correspond to what is in the documentation.
Probably the best solution is to have the documentation pointing to this example but I don't know how to do that.
Thanks

tisonkun:
> because it is the only one I tested

My suggestion is that if you can share how you test it. I'm not familiar with this part so I cannot confidently merging your patch without a technical review from other committers or find my time to verify it. You can wait for that happens or share how you test it so that I'm confident that your patch is correct.

PR #541, opened 2023-04-26
new docs for "Support for launching Pulsar Functions using configuration files"

tpiperatgod:
### Documentation

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

PR #544, opened 2023-04-27
[improve][fn] add docs for supporting reading config options from files in Python runner

Anonymitaet:
This PR adds docs for https://github.com/apache/pulsar/pull/18951. It is based on the tech inputs provided by @tpiperatgod in https://github.com/apache/pulsar-site/pull/541

- [x] `doc` 

cc @D-2-Ed for editorial review, thanks!


Anonymitaet:
Hi @tpiperatgod, I submit this PR since it's more convenient to show you my thoughts and preview. 

### Preview

image

### Doc changes explanations

2️⃣3️⃣ in the image above
I've restricted the doc structure by "showing the example first" and then "explaining the details". 

Reasons:

1) Most developers are likely "learning by doing" rather than "learning by reading". Examples always attract them the most. "Hook them up" first and then they have more patience to read it through.

2) The doc structure is more clear. The beginning is the example, and the latter is the explanation. Splitting them up can increase readability.

1️⃣ 

Consolidate the note into running text to accommodate the new structure.

I've also made other changes like updating all future tense to present tense since "[Write in the simple present tense as much as possible if you are covering facts that were, are, and forever shall be true](https://docs.google.com/document/d/1lc5j4RtuLIzlEYCBo97AC8-U_3Erzs_lxpkDuseU0n4/edit#bookmark=id.e8uqh1awkcnp)" in technical writing.

### Question

4️⃣ 

Seems that the explanations in this box is contradictory.

The above indicates "--config-file" override "command-line", but the below says the opposite. Can you clarify a bit and give some examples? Thank you!

tpiperatgod:
lgtm

And for question 4:

I made a mistake in the first sentence, it should be that the parameters passed in from the command line have a higher priority.

Thank you for the review and the advice on technical writing!

Anonymitaet:
@tpiperatgod thanks for your quick response!

> it should be that the parameters passed in from the command line have a higher priority.

Does that mean I can delete the 1st sentence and only keep the 2nd one?

tpiperatgod:
Yes

Anonymitaet:
> Yes

Done, PTAL.

Anonymitaet:
@momo-jun @Huanli-Meng could you please help review? Thank you!

docs/functions-runtime-process.md
```diff
@@ -22,3 +22,37 @@ functionRuntimeFactoryConfigs:
...
+**Example**
+
+You can start a Python runtime using the configuration file `config.ini` with the following command.
+
+```shell
+pulsar-admin functions localrun \
+  --py /path/to/python_instance.py \
+  --config-file /path/to/config.ini \
+  --classname MyFunction \
+  --logging_leve debug \
```
momo-jun, line 39:
```suggestion
  --logging_level debug \
```

docs/functions-runtime-process.md
```diff
@@ -22,3 +22,37 @@ functionRuntimeFactoryConfigs:
...
+    **Example**
+
+    ```ini
+    [DEFAULT]
+    logging_level = info
+    max_pending_async_requests = 1000
+    max_concurrent_requests = 50
+    ```
+
+- When starting with the ` --config-file`, all parameters can be set in the ` --config-file`. If you specify parameters on the command line that are also present in the ` --config-file`, the values of the parameters on the command line will **take precedence over** the values in the ` --config-file`. Like in the example above, the `logging_level` is set to `debug`.
```
momo-jun, line 58:
```suggestion
- When you set a parameter through both the configuration file and the command line, like `logging_level` in the example above, the value set through the command line will **take precedence over** the one set through the configuration file. As a result, the value of `logging_level` is `debug`.
```

momo-jun (approved):
LGTM in general, with a couple of suggestions.

docs/functions-runtime-process.md
Anonymitaet: Oops! Thank you!

tisonkun:
[site2/website-next/yarn.lock](https://github.com/apache/pulsar-site/pull/544/files#diff-fbcfe767eb3f8738ce6071e1fed0399a302c9fb02cdd67ef1dceeef79200956f) is checked in unintentionally. Reverting...

Anonymitaet:
@tisonkun Oops! Thank you!

PR #580, opened 2023-05-17
[doc] Add doc for openIDKeyIdCacheMissRefreshSeconds setting

michaeljmarshall:
This PR adds doc for https://github.com/apache/pulsar/pull/20338

- [x] `doc`

I rendered the changes using `./preview.sh` and they appear correct.


docs/security-openid-connect.md
```diff
@@ -72,6 +76,10 @@ When using OIDC for a client connecting through the proxy to the broker, it is n
...
...OpenID Connect Discovery 1.0](https://openid.net/specs/openid-connect-discovery-1_0.html) spec gives the `AuthenticationProviderOpenID` a way to discover trusted public keys. The public keys are formatted as a [JSON Web Key (JWK)](https://www.rfc-editor.org/rfc/rfc7517) set, also known as a JWKS. When the Identity Provider rotates signing keys, there is a chance that the Identity Provider will start signing tokens with the new key before the JWKS cache has been refreshed. In order to avoid rejecting tokens signed with the new key, the OIDC Authentication Provider will attempt to refresh the JWKS when a token has a trusted issuer claim but the key ID (kid claim) is not in the issuer's cached JWKS. The `openIDKeyIdCacheMissRefreshSeconds` setting determines how long the OIDC Authentication Provider will wait before attempting to refresh the JWKS. The default value is 300 seconds and means that a JWKS must have been in the cache for at least 300 seconds before a missing key ID will trigger cache invalidation. The `openIDKeyIdCacheMissRefreshSeconds` setting protects the OIDC Authentication Provider from a malicious client that presents a token with a new key ID every time it connects.
```
Anonymitaet, line 81:
```suggestion
The [OpenID Connect Discovery 1.0](https://openid.net/specs/openid-connect-discovery-1_0.html) spec gives the `AuthenticationProviderOpenID` a way to discover trusted public keys. The public keys are formatted as a [JSON Web Key (JWK)](https://www.rfc-editor.org/rfc/rfc7517) set, also known as a JWKS. When the Identity Provider rotates signing keys, there is a chance that the Identity Provider will start signing tokens with the new key before the JWKS cache has been refreshed. To avoid rejecting tokens signed with the new key, the OIDC Authentication Provider will attempt to refresh the JWKS when a token has a trusted issuer claim but the key ID (kid claim) is not in the issuer's cached JWKS. The `openIDKeyIdCacheMissRefreshSeconds` setting determines how long the OIDC Authentication Provider will wait before attempting to refresh the JWKS. The default value is 300 seconds. It means that a JWKS must have been in the cache for at least 300 seconds before a missing key ID triggers cache invalidation. The `openIDKeyIdCacheMissRefreshSeconds` setting protects the OIDC Authentication Provider from a malicious client that presents a token with a new key ID every time it connects.
```

Anonymitaet (commented):
Is this feature available from 3.0? If so, do we need to apply the same doc change to the 3.0.x doc set?

michaeljmarshall:
@Anonymitaet - these docs will be suitable for 3.0.1, but not 3.0.0, which is one weakness of the approach for the 3.0.x scheme. How do you propose we handle this discrepancy? Thanks!

tisonkun:
@michaeljmarshall add a note to specify that it depends on patch release version.

Anonymitaet:
> @michaeljmarshall add a note to specify that it depends on patch release version.

+1, for example,

> ::: tip
> 
> This feature is available from Pulsar 3.0.1.
> 
> :::

PR #590, opened 2023-05-28
chore: remove unnecessary release stuff

tisonkun:
Python client has its own release guide and scripts - https://github.com/apache/pulsar-client-python/blob/main/RELEASE.md

C++ client has its own release guide but still depends on `cpp-apidoc-generator.py` so I keep the related scripts here.

---

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`


tisonkun:
Merging...

cc @BewareMyPower @shibd in case anything that can affect you.

PR #591, opened 2023-05-29
In order to ensure that the api documentation is up-to-date in search…

shy-share:
This PR fixes https://github.com/apache/pulsar/pull/20390

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/static/robots.txt
+++ b/static/robots.txt
@@ -1,5 +1,5 @@
 Sitemap: https://pulsar.apache.org/sitemap.xml
 User-agent: *
-Allow: /api/
-Disallow: /
+Allow: /
+Disallow: /api/
 Crawl-delay: 10
--- a/static/robots.txt
+++ b/static/robots.txt
@@ -1,1 +1,4 @@
-Sitemap: https://pulsar.apache.org/sitemap.xml
+User-agent: *
+Allow: /api
+Disallow: /
+Crawl-delay: 10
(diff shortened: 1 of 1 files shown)


shy-share:
@tisonkun cc

tisonkun (changes requested):
1. Why do you delete the sitemap entry?
2. Allow and Disallow seems reversed.

shy-share:
> 1. Why do you delete the sitemap entry?
> 2. Allow and Disallow seems reversed.

I have already edited it

shy-share:
> 1. Why do you delete the sitemap entry?
> 2. Allow and Disallow seems reversed.

@tisonkun 
image

tisonkun (changes requested):
The goal is that we want legacy API docs not to be crawled. Please check the motivation and target,

shy-share:
> The goal is that we want legacy API docs not to be crawled. Please check the motivation and target,

Well, I understand the meaning is just the opposite

Anonymitaet:
> Well, I understand the meaning is just the opposite

We want to show the latest stable API doc version (e.g., 3.0.x) instead of legacy ones.

static/robots.txt
```diff
@@ -1 +1,5 @@
 Sitemap: https://pulsar.apache.org/sitemap.xml
+User-agent: *
+Allow: /
+Disallow: /api/
```
tisonkun, line 4: We should exclude legacy versions only. Maintained versions should be still indexed.

shy-share:
#623  has solved this problem, so close this pr

PR #408, opened 2023-02-14
Document connecting through a proxy

michaeljmarshall:
### Documentation

Adding docs for https://github.com/apache/pulsar/pull/19455

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

PR #593, opened 2023-05-30
Sync doc updates for proxyRole to versioned docs

momo-jun:
The code PR https://github.com/apache/pulsar/pull/19455 has been applied to corresponding patch versions of Pulsar.

So this PR applies incremental doc updates to the 2.8.x/2.9.x/2.10.x/2.11.x docs (original doc PR: [#408](https://github.com/apache/pulsar-site/pull/408)).

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/versioned_docs/version-2.10.x/security-authorization.md
+++ b/versioned_docs/version-2.10.x/security-authorization.md
@@ -38,12 +38,6 @@ By default, the broker treats the connection between a proxy and the broker as a
 
 Pulsar uses *Proxy roles* to enable the authentication. Proxy roles are specified in the broker configuration file, [`conf/broker.conf`](reference-configuration.md#broker). If a client that is authenticated with a broker is one of its ```proxyRoles```, all requests from that client must also carry information about the role of the client that is authenticated with the proxy. This information is called the *original principal*. If the *original principal* is absent, the client is not able to access anything.
 
-:::note
-
-Starting from 2.10.4, if a Proxy is not correctly configured to use a role that is in the `proxyRoles`, the connection will get rejected.
-
-:::
-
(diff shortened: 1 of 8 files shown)


versioned_docs/version-2.10.x/security-authorization.md
```diff
@@ -38,6 +38,12 @@ By default, the broker treats the connection between a proxy and the broker as a
 
 Pulsar uses *Proxy roles* to enable the authentication. Proxy roles are specified in the broker configuration file, [`conf/broker.conf`](reference-configuration.md#broker). If a client that is authenticated with a broker is one of its ```proxyRoles```, all requests from that client must also carry information about the role of the client that is authenticated with the proxy. This information is called the *original principal*. If the *original principal* is absent, the client is not able to access anything.
 
+:::note
+
+Starting from 2.10.4, if a Proxy is not correctly configured to use a role that is in the `proxyRoles`, the connection will get rejected.
```
michaeljmarshall, line 43: This requirement was only introduced in 3.0.0. It isn't present for any prior versions.

michaeljmarshall (commented):
Thanks for your contribution @momo-jun. The protocol doc is valid, but we don't need to copy the `note`.

versioned_docs/version-2.10.x/security-authorization.md
momo-jun: Oh, I thought the changes were all applied to those patch version branches. I will revert them.

Anonymitaet (approved):

michaeljmarshall (approved):
LGTM

PR #596, opened 2023-06-02
[fix] remove extra dot to fix render problem

labuladong:
This PR fixes https://github.com/apache/pulsar/issues/20470

Just a global replacement to fix this issue. Local preview:

image

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/docs/io-aerospike-sink.md
+++ b/docs/io-aerospike-sink.md
@@ -8,7 +8,7 @@ sidebar_label: "Aerospike sink connector"
 
 You can download all the Pulsar connectors on [download page](pathname:///download).
 
-::::
+:::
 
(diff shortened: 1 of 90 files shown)


tisonkun (approved):
Thank you!

tisonkun:
As it touches a large range of versions, it may take a bit longer time to build. But let's wait a bit to ensure that we don't introduce new syntax error.

tisonkun:
No. It touches only two versions. I'll merge this patch now.

@labuladong do you check that other old version won't be affected?

labuladong:
Other older versions won't be affected, because the older versions don't have the `:::note` tab.

PR #598, opened 2023-06-02
Fix error metric prefix

Technoboy-:
### Motivation

We will convert the metric prefix `brk_` to `pulsar_` when generating the metrics.
So we don't have `brk_` related metrics.

As most of the users are using Pulsar-2.8 at least, so I just modify the error metrics from 2.8~now

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`


Anonymitaet:
@coderzc 
Could you please review this PR from a technical perspective? Then writers will review it from a technical writing side. Thank you!

coderzc (approved):

Anonymitaet (approved):

PR #615, opened 2023-06-20
[fix][doc] fix pulsar consumer name typo 

allemas:
This PR fixes documentation typo.

- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`


tisonkun (approved):
Thank you! Would you check for other versions also so that we keep docs in sync among all the versions?

PR #625, opened 2023-06-28
[improve][doc] Subscription replication doesn't replicate the type

AlvaroStream:
There can be a corner case where 2 consumer subscribe with different type of subscription on each cluster.

The only thing replicated is information about the subscription. Not the type.

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/versioned_docs/version-3.0.x/administration-geo.md
+++ b/versioned_docs/version-3.0.x/administration-geo.md
@@ -229,7 +229,7 @@ If you want to use replicated subscriptions in Pulsar:
   ```
 :::note
 
-* Replication type is not replicated between clusters. However, it's determined by the first consumer that connects to the subscription.
+* Subscription type is not replicated between clusters. However, it's determined by the first consumer that connects to the subscription.
 
 :::
 
--- a/versioned_docs/version-3.0.x/administration-geo.md
+++ b/versioned_docs/version-3.0.x/administration-geo.md
@@ -229,7 +229,7 @@ If you want to use replicated subscriptions in Pulsar:
   ```
 :::note
 
-* Replication type is not replicated between clusters but determined by the first consumer that connects to the subscription.
+* Replication type is not replicated between clusters. However, it's determined by the first consumer that connects to the subscription.
 
 :::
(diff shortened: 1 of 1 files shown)


Anonymitaet:
@gaoran10 
Could you please help review this PR from a technical perspective? Thank you!

gaoran10 (approved):
LGTM. We can find that the broker only creates a subscription after receiving the subscription update marker(refer to [here](https://github.com/apache/pulsar/blob/v3.0.0/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/ReplicatedSubscriptionsController.java#L197)), the subscription type only depends on the first added consumer.

versioned_docs/version-3.0.x/administration-geo.md
```diff
@@ -227,6 +227,12 @@ If you want to use replicated subscriptions in Pulsar:
               .replicateSubscriptionState(true)
               .subscribe();
   ```
+:::note
+
+* Replication type is not replicated between clusters but determined by the first consumer that connects to the subscription.
```
Anonymitaet, line 232:
```suggestion
* Replication type is not replicated between clusters. However, it's determined by the first consumer that connects to the subscription.
```
Can we break this long sentence to improve readability?

AlvaroStream: yes

versioned_docs/version-3.0.x/administration-geo.md
```diff
@@ -227,6 +227,12 @@ If you want to use replicated subscriptions in Pulsar:
               .replicateSubscriptionState(true)
               .subscribe();
   ```
+:::note
+
+* Replication type is not replicated between clusters. However, it's determined by the first consumer that connects to the subscription.
```
hangc0276, line 232: Subscription Type?

AlvaroStream: yes

PR #633, opened 2023-07-07
[improve][doc] clarify detailed behaviors for failover subscription type

Anonymitaet:
This PR fixes https://github.com/apache/pulsar/issues/20751

@codelipenghui many thanks for your technical guidance! 

- [x] `doc`


Anonymitaet:
Previews look good:

image

image

image

image

static/assets/pulsar-failover-subscriptions-5.png
Anonymitaet, line 1: Known issue: this image style (drawn in Figma) does not match the style of other images (drawn in LucidChart) because we can't log in LucidChart suddenly. Will align the image style once we can log in LucidChart.

Anonymitaet:
Known issue 2: we can further enhance this doc by addressing the issue mentioned in https://github.com/apache/pulsar/issues/20752. As that requires engineers to dedicate time to providing technical inputs, I have submitted this PR first.

Anonymitaet:
@codelipenghui @Girish Sharma (sorry I do not find your handle on GitHub) 
Could you please help review this PR? Thank you!

grssam:
Thank you @Anonymitaet for the quick update on docs.

In my organization, I've internally also emphasized on the effect of a new consumer joining the subscription and the current state of preservation of one-consumer-per-partition boundaries keeping ordered consumption in mind.
Pasting some excerpt here which may be helpful:




> As you add more consumers consuming from the topic, each new consumer leads to partition re-balancing. You can expect this bahaviour as part of your application rollout, incremental pod restarts, etc. Once the consumer count becomes stable, rebalancing stops.

> Consider a case when you have 9 partitions in a topic and you are starting up your application. Initially, you would have lower consumer count and distribution would be like:

> (image for the case where consumer < partition)

> In this approach, no consumer is siting idle. Even if any of the 3 consumers go down, the partitions would be rebalanced across the remaining ones.

> As you add more consumers (more of your application pods come up), You would end up in the following situation:

> (image for the case where consumer > partition)

> Here, Consumer B and D are sitting idle and not receiving any messages until one of A and C go down.

> `(as a warning note)` Unlike KEY_SHARED subscription, there is no checkpointing done during the consumer switch (read below). Thus, there can be cases where for a partition, the older consumer (if still active) is still processing some messages while the newly switched over active consumer starts getting new messages. This may result in duplicate messages, or in worst case, out of order processing of few messages.


I've also provided a recipe in my documentation about how to have more deterministic consumers and avoid re-balancing for most cases, but that may be out of scope for pulsar site. Let me know.

Anonymitaet:
Hi @grssam thanks for your guidance! I've discussed this with @codelipenghui and added explanations in https://github.com/apache/pulsar-site/pull/633/commits/cc9aa879d494132b5a4fee00cf771c46fd2029ef, does this LGTY?


Besides, this is another independent topic (like a best practice in a special use case) that is out of the scope of this PR. How about recording this doc request on Pulsar GitHub first and see if we can add it to Pulsar docs later?

> I've also provided a recipe in my documentation about how to have more deterministic consumers and avoid re-balancing for most cases, but that may be out of scope for pulsar site. Let me know.

grssam:
> Hi @grssam thanks for your guidance! I've discussed this with @codelipenghui and added explanations in [cc9aa87](https://github.com/apache/pulsar-site/commit/cc9aa879d494132b5a4fee00cf771c46fd2029ef), does this LGTY?

Looks good.

> Besides, this is another independent topic (like a best practice in a special use case) that is out of the scope of this PR. How about recording this doc request on Pulsar GitHub first and see if we can add it to Pulsar docs later?

Makes sense. Could you help me with the process for doc request for pulsar github?

Anonymitaet:
@grssam Thanks! 

1. Can you approve this PR if you have no more comments? Then we can merge it

3. You can open a doc issue at [here ](https://github.com/apache/pulsar/issues/new?assignees=&labels=doc-required&projects=&template=doc.yml&title=%5BDoc%5D+) with context

> Could you help me with the process for doc request for pulsar github?

grssam (approved):

docs/concepts-messaging.md
```diff
@@ -521,6 +521,12 @@ A master consumer is picked for a non-partitioned topic or each partition of a p
 
 When the master consumer disconnects, all (non-acknowledged and subsequent) messages are delivered to the next consumer in line.
 
+::: note
+
+In some cases, a partition may have an older active consumer processing messages while a newly switched over active consumer starts receiving new messages. This may lead to message duplication or out-of-order.
+
+:::
+
```
grssam, line 529: this note is not rendering correctly on the website

Anonymitaet: Good catch! I'll fix it. Thanks!

PR #651, opened 2023-07-21
[improve] [doc] Add compatibility reminders for retry queues

poorbarcode:
This PR fixes https://github.com/apache/pulsar/issues/20612

The default naming rule of Retry topic is `{tenant}/{namespaces}/{topic}-{sub}-RETRY`, see: https://pulsar.apache.org/docs/3.0.x/concepts-messaging/#retry-letter-topic

The rule before version `2.8` was `{tenant}/{namespaces}/{sub}-RETRY`, and if there is a ZK node `/admin/partitioned-topics/{namespace}-RETRY`, Pulsar will still use the rule `{tenant}/{namespaces}/{sub}-RETRY` even if the version of Pulsar is larger than `2.8`.  This part of the explanation is not reflected in the documentation.

**Related PR**
- [#6449 Support Consumers Set Custom Retry Delay](https://github.com/apache/pulsar/pull/6449)
- [#10129 fix the default retry letter and dead letter topic name](https://github.com/apache/pulsar/pull/10129)

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/docs/concepts-messaging.md
+++ b/docs/concepts-messaging.md
@@ -244,12 +244,9 @@ The default retry letter topic uses this format:
 <topicname>-<subscriptionname>-RETRY
 ```
 
-Note: In versions `2.6.x` and `2.7.x`, the default retry letter topic uses this format is `<subscriptionname>-RETRY`.
-If you are upgrading from version `2.6.x~2.7.x` to version `2.8.x` or greater, then you need to manually delete
-the historical retry letter topics and retry letter partitioned topics. Otherwise, Pulsar will continue to use the
-original topics, which were formatted with `<subscriptionname>-RETRY`. Rule `<subscriptionname>-RETRY` should be avoided
-to use, because if multiple topics under the same namespace have the same subscription, the retry message topic names
-for multiple topics will be the same, resulting in mutual consumption.
+:::note
+- For Pulsar 2.6.x and 2.7.x, the default retry letter topic uses the format of `<subscriptionname>-RETRY`. If you upgrade from 2.6.x~2.7.x to 2.8.x or later, you need to delete historical retry letter topics and retry letter partitioned topics. Otherwise, Pulsar continues to use original topics, which are formatted with `<subscriptionname>-RETRY`.
+- It is not recommended to use `<subscriptionname>-RETRY` because if multiple topics under the same namespace have the same subscription, then retry message topic names for multiple topics might be the same, which will result in mutual consumptions.
 
 Use the Java client to specify the name of the retry letter topic.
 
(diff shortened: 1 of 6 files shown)


Anonymitaet:
@gaoran10 
Could you please help review this PR from a technical perspective? Then writers will review it from a technical writing side. Thank you!

Anonymitaet:
ping @gaoran10 to review, tia!

Anonymitaet:
ping @gaoran10 to review, tia!

xiaogang-sn:
If manually specifying the name of retry letter topic or dead letter topic can also achieve the goal we can mention it along the changes as well.

It may also be worthy noting that using `<subscriptionname>-RETRY` as topic name may cause trouble when multiple topics in the same namespace having the subscriptions with the same name.

poorbarcode:
@xiaogang-sn 

> It may also be worthy noting that using <subscriptionname>-RETRY as topic name may cause trouble when multiple topics in the same namespace having the subscriptions with the same name.

Correct, so in `2.8.x`, the format rule was changed to `{topic}-{sub}-RETRY`

xiaogang-sn:
I would suggest to add the consequence in brief if continuing using `<subscriptionname>-RETRY`as retry topic name or add a reference link to the relevant issue/pr.

xiaogang-sn (approved):

docs/concepts-messaging.md
```diff
@@ -244,6 +244,13 @@ The default retry letter topic uses this format:
 <topicname>-<subscriptionname>-RETRY
 ```
 
+Note: In versions `2.6.x` and `2.7.x`, the default retry letter topic uses this format is `<subscriptionname>-RETRY`.
+If you are upgrading from version `2.6.x~2.7.x` to version `2.8.x` or greater, then you need to manually delete
+the historical retry letter topics and retry letter partitioned topics. Otherwise, Pulsar will continue to use the
+original topics, which were formatted with `<subscriptionname>-RETRY`. Rule `<subscriptionname>-RETRY` should be avoided
+to use, because if multiple topics under the same namespace have the same subscription, the retry message topic names
+for multiple topics will be the same, resulting in mutual consumption.
```
Anonymitaet, line 252:
```suggestion
:::note

- For Pulsar 2.6.x and 2.7.x, the default retry letter topic uses the format of `<subscriptionname>-RETRY`. If you upgrade from 2.6.x~2.7.x to 2.8.x or later, you need to delete historical retry letter topics and retry letter partitioned topics. Otherwise, Pulsar continues to use original topics, which are formatted with `<subscriptionname>-RETRY`. 

- It is not recommended to use `<subscriptionname>-RETRY` because if multiple topics under the same namespace have the same subscription, then retry message topic names for multiple topics might be the same, which will result in mutual consumptions.

:::
```

poorbarcode: Fixed

poorbarcode: 截屏2023-07-28 11 15 29

gaoran10 (approved):
LGTM

Anonymitaet (approved):

PR #708, opened 2023-09-26
[fix][doc]Update security-encryption.md for key pairs of golang

nineep:
golang client: only RSA keys are supported

func (d *DefaultMessageCrypto) addPublicKeyCipher(keyName string, keyReader KeyReader) error {
	d.cipherLock.Lock()
	defer d.cipherLock.Unlock()
	if keyName == "" || keyReader == nil {
		return fmt.Errorf("keyname or keyreader is null")
	}

	// read the public key and its info using keyReader
	keyInfo, err := keyReader.PublicKey(keyName, nil)
	if err != nil {
		return err
	}

	parsedKey, err := d.loadPublicKey(keyInfo.Key())
	if err != nil {
		return err
	}

	// try to cast to RSA key
	rsaPubKey, ok := parsedKey.(*rsa.PublicKey)
	if !ok {
		return fmt.Errorf("only RSA keys are supported")
	}

	encryptedDataKey, err := rsa.EncryptOAEP(sha1.New(), rand.Reader, rsaPubKey, d.dataKey, nil)
	if err != nil {
		return err
	}

	d.encryptedDataKeyMap.Store(keyName, NewEncryptionKeyInfo(keyName, encryptedDataKey, keyInfo.Metadata()))

	return nil
}

This PR adds doc for #xyz

This PR fixes #xyz 

- [x] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/versioned_docs/version-3.1.x/security-encryption.md
+++ b/versioned_docs/version-3.1.x/security-encryption.md
@@ -51,7 +51,7 @@ If the produced messages are consumed across application boundaries, you need to
    ````mdx-code-block
    <Tabs groupId="lang-choice"
      defaultValue="ECDSA"
-     values={[{"label":"ECDSA (for Java and Go clients)","value":"ECDSA"},{"label":"RSA (for Python, C++ and Node.js clients)","value":"RSA"}]}>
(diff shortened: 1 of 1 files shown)


tisonkun:
cc @nodece May you do a technical review for this change?

PR #750, opened 2023-12-13
Improve GPG instructions in the contribution guide

lhotari:
The GPG instructions are unclear and some instructions (such as MacOS keychain integration) are missing.
This PR improves the instructions.

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/contribute/create-gpg-keys.md
+++ b/contribute/create-gpg-keys.md
@@ -15,7 +15,7 @@ brew install gnupg
 brew install pinentry-mac
 mkdir ~/.gnupg
 chmod 0700 ~/.gnupg
-echo "pinentry-program $(brew --prefix)/bin/pinentry-mac" | tee ~/.gnupg/gpg-agent.conf
+echo "pinentry-program $(brew --prefix)/bin/pinentry-mac" | tee -a ~/.gnupg/gpg-agent.conf
 ```
 
 Set configuration to use `SHA512` keys by default:
--- a/contribute/create-gpg-keys.md
+++ b/contribute/create-gpg-keys.md
@@ -11,6 +11,11 @@ Install GnuPG. For example on macOS:
 
 ```shell
 brew install gnupg
+# On MacOS, install keychain integration
+brew install pinentry-mac
+mkdir ~/.gnupg
+chmod 0700 ~/.gnupg
+echo "pinentry-program $(brew --prefix)/bin/pinentry-mac" | tee ~/.gnupg/gpg-agent.conf
 ```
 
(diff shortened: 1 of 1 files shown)


contribute/create-gpg-keys.md
```diff
@@ -78,19 +92,29 @@ Change (N)ame, (C)omment, (E)mail or (O)kay/(Q)uit? O
...
+Use the key id to publish it to several public key servers:
 
-:::
+```shell
+# find out your key id
+APACHEID=your_asf_id
+KEY_ID=$(gpg --list-keys --with-colons $APACHEID@apache.org | egrep "^pub" | awk -F: '{print $5}')
+echo "key id: $KEY_ID"
+# send the public key to multiple servers
+gpg --send-key $KEY_ID
```
onobc, line 105: These sends fail w/ `"no route to host"`. Instead the documentation says to use 
```
gpg --send-key --keyserver=hkp://pool.sks-keyservers.net $KEY_ID
```
Although that resulted in a "server indicated a failure" response. However, it looks like the key was added as I got an email indicating so. I think we can add the following verbiage to the end of this step. 


<NEW_CONTENT>
If the `send-key` steps fail with `no route to host` then instead try the following command:

```
gpg --send-key --keyserver=hkp://ipv4.pool.sks-keyservers.net $KEY_ID
```
:::note The above command may return a `"gpg: keyserver send failed: Server indicated a failure"` response but most likely the key was added. Check your email for a message from keyserver@keys.openpgp.org with subject `"Your key upload on keys.openpgp.org"` and verify the key in the email body matches your submitted key. If you do not receive an email, reach out to another member for help.

</NEW_CONTENT>

Wdyt?

lhotari: The `ipv4.pool.sks-keyservers.net` address no longer exists. Doing `dig ipv4.pool.sks-keyservers.net` or `nslookup ipv4.pool.sks-keyservers.net` will fail.

I think that the problem could actually be related to DNS. 
I found this resolution in StackOverflow:
https://stackoverflow.com/a/67259857

I think it might be the proper resolution.

onobc: What is concerning is that it seems that some of the servers are going away soon? Is there some official docs that we could instead point to that says "If you have issues please see the offiical docs..."

onobc: I just saw your commit. If that works then this is sufficient. Thanks for the update

onobc (approved):
LGTM - I verified the steps when adding my new key

PR #807, opened 2024-02-29
Add the "I don't understand it" link

visortelle:
It's rather an idea.

Screenshot 2024-03-01 at 12 11 14 AM

**Goals:**
- Allow users to quickly ask a question in the right place.
- Improve the documentation by encouraging users to report unclear or incomplete parts of the documentation.

Try it live: https://pulsar-site-2024-updates-7uwroc4qs-visortelle.vercel.app/docs/next/concepts-cluster-level-failover/

If there are too many questions, we can change the link to another Google-indexable place like https://github.com/apache/pulsar/discussions/categories/q-a

Alternatively, we could add more links to doc page's bottom section that encourages users to communicate with Pulsar developers and the community:

Screenshot 2024-03-01 at 12 36 57 AM

https://docs.confluent.io/platform/current/clients/index.html

### DIFF
--- a/src/components/ui/ArticleThumb/ArticleThumb.js
+++ b/src/components/ui/ArticleThumb/ArticleThumb.js
@@ -51,8 +51,8 @@ const ArticleThumb = () => {
       </div>
       <a
         target='_blank'
-        className={s.RequestImprovementButton}
-        href="https://github.com/apache/pulsar/issues/new?assignees=&labels=doc-required&projects=&template=doc.yml&title=%5BDoc%5D+"
+        className={s.RequestHelpButton}
+        href="https://github.com/apache/pulsar/discussions/new?category=q-a"
       >
         🛟 I don't understand it
       </a>
--- a/src/components/ui/ArticleThumb/ArticleThumb.js
+++ b/src/components/ui/ArticleThumb/ArticleThumb.js
@@ -1,56 +1,63 @@
-import React, {useState} from 'react';
+import React, { useState } from 'react';
 import ThumbDown from './thumbs-down.svg';
 import ThumbUp from './thumbs-up.svg';
+import s from './ArticleThumb.module.css';
 
 const ArticleThumb = () => {
-    const [activeKey, setActiveKey] = useState();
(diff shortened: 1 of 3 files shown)


visortelle:
Changed the button link to https://github.com/apache/pulsar/discussions/categories/q-a

Merging. Let's see if this works 🙂

asafm:
1. I don't think we should merge without review. 
2. We need to improve, but in the current state of the community, who will take this pile of work and fix it? 
3. This kind of duplicates the thumbs up and down above. Maybe just this button instead of the thumbs up or down. Where do the analytics for the thumbs go?

visortelle:
**1**

> I don't think we should merge without review.
Answered here: https://github.com/apache/pulsar-site/pull/809#issuecomment-1976922341

**2**

> We need to improve, but in the current state of the community, who will take this pile of work and fix it?

I'm not sure that I have an answer to this question.

**3**

> This kind of duplicates the thumbs up and down above. Maybe just this button instead of the thumbs up or down. Where do the analytics for the thumbs go?

Agree here. A few ups/downs without any meaningful feedback don't make much sense unless there are hundreds of these ups/downs per page.

asafm:
@lhotari Do you have any idea about the thumbs down/up analytics?

asafm:
> It's rather a rhetorical question.

I'm serious. You can create a gazillion of issues, but nobody exists, today, that will actually do it.
My thinking  - still pondering it - is once we finish this huge epic (pip-249) of the site rework - which is just around the corner, start the second big project I had in mind - reworking the documentation site completely:
- I have the blueprint, which is a big table of contents. 
- I have a more detailed design for the first section: Getting Started. Laid out in an approved PIP which I haven't executed yet. https://github.com/apache/pulsar/issues/19912

This project needs at least 2-3 people - can't do it alone. So I still need to think this through, to see if it makes sense.

visortelle:
> This project needs at least 2-3 people - can't do it alone.

I'm ready to help with the big docs project, but I can't guarantee full involvement.
It all depends on the time availability and other resources.

PR #825, opened 2024-03-06
The sample configuration of io-debezium has an error of parameter.

zhaoyajun2009:
This PR adds doc for #xyz

This PR fixes #xyz

### DIFF
--- a/docs/io-cdc-debezium.md
+++ b/docs/io-cdc-debezium.md
@@ -80,7 +80,7 @@ You can use one of the following methods to create a configuration file.
   tenant: "public"
   namespace: "default"
   name: "debezium-mysql-source"
-  inputs: [ "debezium-mysql-topic" ]
+  topicName: "debezium-mysql-topic"
   archive: "connectors/pulsar-io-debezium-mysql-@pulsar:version@.nar"
   parallelism: 1
 
@@ -378,7 +378,7 @@ You need to create a configuration file before using the Pulsar Debezium connect
   tenant: "public"
   namespace: "default"
   name: "debezium-mongodb-source"
-  inputs: [ "debezium-mongodb-topic" ]
+  topicName: "debezium-mongodb-topic"
   archive: "connectors/pulsar-io-debezium-mongodb-@pulsar:version@.nar"
   parallelism: 1
 
--- a/docs/admin-api-schemas.md
+++ b/docs/admin-api-schemas.md
@@ -202,7 +202,7 @@ To get a specific version of a schema, you can use one of the following methods.
 Use the `get` subcommand.
 
 ```bash
-pulsar-admin schemas get <topic-name> --version <version>
+pulsar-admin schemas get <topic-name> --version=<version>
 ```
 
 </TabItem>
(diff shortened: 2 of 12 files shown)


visortelle:
I suppose the PR author is right here: https://github.com/apache/pulsar/blob/e2f94dc98dbecb4dc401ba837c54f497ca9d896f/pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/io/SourceConfig.java#L43
 
But it would be good if someone else took a look.

docs/io-cdc-debezium.md
```diff
@@ -80,7 +80,7 @@ You can use one of the following methods to create a configuration file.
   tenant: "public"
   namespace: "default"
   name: "debezium-mysql-source"
-  inputs: [ "debezium-mysql-topic" ]
+  topicName: "debezium-mysql-topic"
```
shibd, line 83: In fact, the CDC Debezium Source connector does not use this configuration, but sends data to the pulsar topic according to this rule.

```
The topic naming rule is: "{{database.server.name}}.{{table.name}}". For examples: "public/default/mydbserver.public.io-test".You can also choose to use a variety of other tools to create a connector:
```

We should remove this line.

zhaoyajun2009: if remove this line，the process of testing is unable to function properly

zhaoyajun2009: if config this topicName，it create the topic as follow：
`persistent://public/default/debezium-mysql-topic3333333333333333333333333333333333-partition-0`
the logs is printed as follow：
微信图片_20240307135342

zhaoyajun2009:
> I suppose the PR author is right here: https://github.com/apache/pulsar/blob/e2f94dc98dbecb4dc401ba837c54f497ca9d896f/pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/io/SourceConfig.java#L43
> 
> Strangely, this mistake went unnoticed for a long time. It would be good if someone else took a look.



> I suppose the PR author is right here: https://github.com/apache/pulsar/blob/e2f94dc98dbecb4dc401ba837c54f497ca9d896f/pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/io/SourceConfig.java#L43
> 
> Strangely, this mistake went unnoticed for a long time. It would be good if someone else took a look.

because the parameter of "inputs" is in the SinkConfig, it is caused by copying.  
https://github.com/apache/pulsar/blob/e2f94dc98dbecb4dc401ba837c54f497ca9d896f/pulsar-client-admin-api/src/main/java/org/apache/pulsar/common/io/SinkConfig.java#L49C39-L50C1

docs/io-cdc-debezium.md
shibd: I remember that I didn't need to fill in this name in my previous test, and you can refer to the [this docs](https://docs.streamnative.io/hub/connector-debezium-postgres-source-v3.1).

whatever, there will be no problem with the configuration.

shibd (approved):

PR #826, opened 2024-03-06
The sample configuration of io-debezium has an error of container's name.

zhaoyajun2009:
The container‘s name created above is “pulsar-postgres”， but the container's name entered below is "pulsar-postgresql".

This PR adds doc for #xyz

This PR fixes #xyz


visortelle:
@zhaoyajun2009 thanks! 🙌

PR #849, opened 2024-03-14
Show supported versions on download page

visortelle:
It probably makes sense to place the release schedule and supported versions in a more visible place to regular users.

Now users should perform the following trip to get this info: Download -> Release Notes -> Release Policy

The Release Policy page is on the `/contribute/release-policy/` route, but it's unlikely that most of the users are contributors.

Also, there is no way to figure out which version is LTS from the `/download` page.

### Proposed Solution

- Show supported versions on the `/download` page

Screenshot 2024-03-14 at 11 16 38 PM

---

Show all versions on the `/contribute/release-policy` (no changes here)

Screenshot 2024-03-14 at 11 01 11 PM

### DIFF
--- a/src/components/SupportedVersionsTable.tsx
+++ b/src/components/SupportedVersionsTable.tsx
@@ -1,5 +1,5 @@
 import lodash from 'lodash'
-import React from 'react'
+import React, { FC } from 'react'
 import semver from "semver/preload"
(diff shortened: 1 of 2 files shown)


lhotari (approved):
LGTM. Great improvement @visortelle

jak78:
@visortelle +100

PR #857, opened 2024-03-19
Fix broken Case Studies page

visortelle:


asafm:
How did it broke?

visortelle:
Array item index numbers were used for React keys and after shuffling that led to applying wrong attributes to DOM elements https://github.com/apache/pulsar-site/pull/860/files#diff-f0b43be5610a8cb248163710fe889d314ca103aea58e38abbf1af3321d8608d0L213

PR #859, opened 2024-03-19
Improve team shuffling code

lhotari:
- avoid reassigning team.pmc and team.committers


lhotari:
obsolete since @visortelle is fixing this

PR #863, opened 2024-03-19
Fix broken anchor link for community section

visortelle:


asafm:
It would help a bit if you can describe briefly what was fixed. I can't extract it from the CSS :)

PR #871, opened 2024-03-29
Recommend ZK for Pulsar standalone

visortelle:
### ✅ Contribution Checklist

---

- Recommend considering using ZooKeeper for Pulsar standalone. RocksDB metadata store sometimes doesn't work after restarting a container: https://github.com/apache/pulsar/discussions/20985#discussioncomment-8951010
- Changed argument order for `chown` command because it didn't work on MacOS


merlimat:
According to the issue it doesn't look to be a RocksDB metadata issue, rather a BK bug

versioned_docs/version-3.2.x/getting-started-docker.md
```diff
@@ -56,6 +56,22 @@ bin/pulsar standalone
 </Tabs>
 ````
 
+:::tip
+
+You may encounter issues with the default RocksDB metadata store after restarting Pulsar standalone.
+
+We recommend you consider using the following environment variable to use ZooKeeper as the metadata store:
```
merlimat, line 63: I don't think it's a good idea to recommend not to use a default setting. If we have a bug in default configuration, we must fix it. If it cannot be fixed, we will have to change the defaults.

visortelle:
@merlimat you're right. I removed the mention of the wrong tracking issue.

I'm a little confused about the different problems I encountered with restarting the Pulsar standalone.

The issue with RocksDB I have:
https://github.com/apache/pulsar/issues/15896#issuecomment-1951427536

`Error generating ledger id: org.apache.pulsar.metadata.api.MetadataStoreException: org.rocksdb.RocksDBException: TimedOut(LockTimeout)`

I have an app that concurrently creates about 15 topics with roughly 50 partitions and creates a single producer and consumer for each topic.

The problem was fixed after I switched to creating resources sequentially.

I can share a docker image with the app if it makes sense.

versioned_docs/version-3.2.x/getting-started-docker.md
visortelle: I would be the best solution. @lhotari 's concern was that:

> it would have to be documented since many might be using old versions and wondering about the resolution.
> The problem in changing the default is that it would break existing "deployments" with data

lhotari:
I made a separate commit 96cdbf2 for fixing the chown

versioned_docs/version-3.2.x/getting-started-docker.md
lhotari: > I don't think it's a good idea to recommend not to use a default setting. If we have a bug in default configuration, we must fix it. If it cannot be fixed, we will have to change the defaults.

@merlimat I agree. Since we cannot change already released versions and users are hitting problems, I think it's worth documenting the workaround until there's a better solution. It seems that there isn't a nice way to solve this problem.

lhotari: I'll merge this PR with the current content. We can come back to this later once there's a better resolution.

lhotari (approved):
LGTM, although as discussed there isn't a good solution and it's better to have the workaround documented.

PR #874, opened 2024-04-02
 [feat][doc] show how to use Pulsar BOM with Java client

onobc:
This code proposal adds a section to the Pulsar Java Client reference docs which shows how to use the [Pulsar BOM](https://github.com/apache/pulsar/blob/master/pip/pip-326.md) when setting up the Java client. The BOM ensures that all Pulsar dependencies (both direct and transitive) are at the same expected version.

Here is a motivating discussion which is the catalyst for this improvement: https://github.com/apache/pulsar/discussions/22328

### ✅ Contribution Checklist


versioned_docs/version-3.2.x/client-libraries-java-setup.md
```diff
@@ -42,10 +42,53 @@ If you use Gradle, add the following information to the `build.gradle` file.
 def pulsarVersion = '@pulsar:version@'
 
 dependencies {
-    compile group: 'org.apache.pulsar', name: 'pulsar-client', version: pulsarVersion
+    implementation "org.apache.pulsar:pulsar-client:${pulsarVersion}"
```
onobc, line 45: **[UNRELATED]** updated the configuration to `implementation` and uses the short form of the dependency.

docs/client-libraries-java-setup.md
```diff
@@ -42,10 +42,53 @@ If you use Gradle, add the following information to the `build.gradle` file.
 def pulsarVersion = '@pulsar:version@'
 
 dependencies {
-    compile group: 'org.apache.pulsar', name: 'pulsar-client', version: pulsarVersion
+	implementation "org.apache.pulsar:pulsar-client:${pulsarVersion}"
```
onobc, line 45: **[UNRELATED]** updated the configuration to `implementation` and uses the short form of the dependency.

versioned_docs/version-3.2.x/client-libraries-java-setup.md
```diff
@@ -42,10 +42,53 @@ If you use Gradle, add the following information to the `build.gradle` file.
...
 }
 ```
 
+### Pulsar BOM
```
onobc, line 49: I did not see an easy way to not violate DRY principle here and factor this section out and use it across main and 3.2x versioned docs. Even if we Markdown supported including other Markdown files (which I dont think it does) we would still have the cross-version to think about. For now it is duplicated in 2 places.

lhotari: Yes, the solution isn't "clean" in that sense.

lhotari (approved):
LGTM. Thanks for the contribution @onobc !

PR #875, opened 2024-04-02
Add best consume practice

liangyepianzhou:
image

### ✅ Contribution Checklist


docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+For case 2, where the business system gets stuck or an error occurs which causes the received message to be missed for
+processing, configuring ack timeout is a good solution. The consumer will record every message received on the client side.
+If these messages have not been acknowledged after the specified time, the consumer will request the broker to resend these
+messages to other brokers.
+**Suggestion:** The ack timeout should be set slightly longer based on the message processing speed of the business system.
+If the business system is still processing messages, but the processing time is too long or the timeout is set too small,
+it may result in duplicate message consumption.
+![img_3.png](../static/img/blog-consume-best-practice/AckTimeout.png)
+```java
+        Consumer<Integer> consumer = pulsarClient.newConsumer(Schema.INT32)
```
visortelle, line 145: Code indentation probably should be fixed here and in the other code blocks. 

Screenshot 2024-04-02 at 9 40 49 AM

liangyepianzhou: Fixed. Thanks for the suggestion.

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+  * **reconsumeLaterCumulative**
+  * **reconsumeLater**
+* **negativeAcknowledge**
+* **redeliverUnacknowledgedMessages**
+
+The **ackTimeout** is an automatic protection mechanism. If a consumer configured ackTimeout, the messages will be
+auto-redelivered when the received messages are not acknowledged after a long time. It works at the client side, ensuring
+the unacknowledged messages will be redelivered to another consumer when the connection of a consumer remains active
+but the business system gets stuck or misses to ack some messages. If the consumer gets disconnected when the client crashes,
+the messages will be auto-redelivered by the broker too. However, this mechanism does not wait for the response of acknowledgment,
```
asafm, line 63: Can you please explain that. I can't understand.

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+so if an acknowledgment fails on the broker side or proxy side, an ack hole may occur.
+
+The **deadLetterPolicy** is a policy in the message queue used to handle messages that cannot be processed properly.
+In many message queue systems (such as RabbitMQ, Google Pub/Sub, Apache Pulsar, etc.), this strategy is implemented.
+In Pulsar, the deadLetterPolicy is implemented at the client side, it creates a new retry letter topic and dead letter
+topic when building a consumer with `deadLetterPolicy` configuration. When a consumer calls `reconsumeLaterCumulative`
+or `reconsumeLater`, the message (method parameter) will be produced to the retry letter topic until the retry time reaches
```
asafm, line 70: Let add:
the message will be acknowledged, and a new message, which is a copy of the message read, will be produced to the retry letter topic, bearing a retry number header. When the retry number exceed `maxRedeliverCount`, the message will be produced instead of the dead-letter topic.
When calling `reconsumerLater(msg, delay)` the message written to the retry letter topic will produced with delivery delay - it will be persisted immediately by the broker, but will only be delivered to the consumer after `delay` has passed. It's perfect to space out retries.

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+the `maxRedeliverCount`. The message will be produced to the dead letter topic when the retry time reaches the `
+maxRedeliverCount`. The main difference between them is that `reconsumeLaterCumulative` will cumulative ack the
```
asafm, line 72: I can not understand why would you use cumulativeAck with retry.

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+message (method parameter) after it is produced and `reconsumeLater` will individual ack the message (method parameter)
+after it is produced.
+
```
asafm, line 75: Add note:
NOTE: TX are not enabled / supported for this. This means the two operations of acknowledging the original message and publishing the new message to the retry topic does not happen atomically. It can happen that the acknowledge of the original message  fails, but the new message is produced, so the consumer will consume the same message twice.

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+The **deadLetterPolicy** is a policy in the message queue used to handle messages that cannot be processed properly.
+In many message queue systems (such as RabbitMQ, Google Pub/Sub, Apache Pulsar, etc.), this strategy is implemented.
+In Pulsar, the deadLetterPolicy is implemented at the client side, it creates a new retry letter topic and dead letter
+topic when building a consumer with `deadLetterPolicy` configuration. When a consumer calls `reconsumeLaterCumulative`
+or `reconsumeLater`, the message (method parameter) will be produced to the retry letter topic until the retry time reaches
+the `maxRedeliverCount`. The message will be produced to the dead letter topic when the retry time reaches the `
+maxRedeliverCount`. The main difference between them is that `reconsumeLaterCumulative` will cumulative ack the
+message (method parameter) after it is produced and `reconsumeLater` will individual ack the message (method parameter)
+after it is produced.
```
asafm, line 74: Let's clarify: 
1. The consumers automatically subscribe to the retry letter topic.
2. It's important that the retry-letter topic name will include both the topic name and the subscription name, since we want to avoid having two subscription reading from the topic, writing their retries into the same retry topic. The retries is a state only in the context of your subscription.

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+
+The `negativeAcknowledge` is used to redeliver certain unacknowledged messages while `redeliverUnacknowledgedMessages`
+is used to redeliver all the unacknowledged messages received by this consumer. The main difference between them and
+deadLetterPolicy is that there is no new topic created, and there is an unlimited number of redeliveries.
+
+## Best Practice Suggestion
+
+Different scenarios require different best practices. Users who value the order of partition messages and wish to batch
+process data should choose or implement an appropriate routerPolicy to send a batch of ordered messages to the same
```
asafm, line 83: You're talking about producer best practices here? If so, I think it's out of scope for Redelivery content, no?

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+partition. They should also select either Exclusive or Failover subscription modes. For users who do not care about
+message order and those in stream processing scenarios, they can opt to use Shared and Key-shared subscription modes.
```
asafm, line 85: do not care about message order *and* those in stream processing scenario (which usually do care about order) - this is a contradiction

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Shared && Key-shared
+#### At-least-once
+The term `At least once` ensures that the message is processed at least once, and there is a risk of duplicate messages,
```
asafm, line 89: This sentence was explained better above in the one I gave IMO.

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+but the performance is better than the `Exactly once` semantics. For the `At least once` semantics in shared or
```
asafm, line 90: You can't ever achieve exactly once for consumption, by only using Pulsar. It relies on your target system providing idempotency.

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+key-shared mode, the most important matter is avoiding ack-hole as much as possible. The ack-hole refers to instances
```
asafm, line 91: I think this all paragraph and numbers below do not add any new essential information.

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+where some single messages are missed for acknowledgment. This can occur in the following cases (All of these are the cases
+where the connection is not interrupted, and reconnection will automatically resend the message.):
+
+1. Consumer receives messages but fails to process due to business system error.
+2. Consumer receives messages but misses to process or the business system gets stuck.
+3. Consumer acknowledges the message, but the acknowledge request is lost when sending to the broker. The possibility of
+4. data loss exists in the case of TCP long connections, although the probability is extremely low.
+   ![img_8.png](../static/img/blog-consume-best-practice/Ack-hole.png)
+
+For case 1, configuring `deadLetterPolicy` is a good solution. When the business system receives a signal of message
```
asafm, line 101: This explanation is redundant IMO. Let's expand at the section above.

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+processing failure, it can immediately check and decide whether to retry after a period of time. After they call
+`reconsumeLater` API, the message will be acknowledged and resent to the retry letter topic that is automatically
+subscribed by the consumer. When it reaches `maxRedeliverCount`, the message will be sent to the dead letter topic.
+Messages sent to the dead letter topic should be considered as non-retryable messages, and we recommend setting an
```
asafm, line 105: This should be specified above.

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+
+For case 3, there are no effective preventive measures. This is because all methods of redelivery are triggered by the
+client when the connection is not disconnected, and the client does not wait for an ack response by default.
+
+`Short-term solution`: This should be an extreme case where users can unload the topic to resolve after observing
+an abnormal ack hole (existing for more than 3 * ack time out).
+`Long-term solution`: Modify the ack-timeout mechanism to wait for the acknowledgment response.
+
+#### Exactly-once
+If the users cannot tolerate message repetition, they can acknowledge messages with a transaction. Transaction can
```
asafm, line 162: I don't understand why. @codelipenghui ?

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+prevent repeated consumption. If a message has been acknowledged, it will wait for a response and throw
+`TransactionConflictException` when the client acknowledges the message with a transaction.
+
+**Notices:** When using transactions, do not configure DeadLetterPolicy, but instead use negativeAcknowledge to resend messages.
```
asafm, line 166: Why?

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+        } catch (Exception e) {
+            if (!(e.getCause() instanceof PulsarClientException.TransactionConflictException)) {
+                consumer.negativeAcknowledge(message);
+            }
+            transaction.abort().get();
+        } finally {
+            message.release();
+        }
+```
+### Exclusive && Failover
```
asafm, line 192: We have to stop here IMO. Something here in the structure makes it way too long IMO.

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+            consumer.acknowledgeCumulative(message);
+        } catch (Exception e) {
+            consumer.redeliverUnacknowledgedMessages();
+        } finally {
+            message.release();
+        }
+````
+
+#### Exactly-once (Beta)
+In the `Exclusive` and `Failover` mode, the most troublesome issue for users is the problem of duplicate messages caused
```
asafm, line 222: Didn't understand anything here.

static/img/blog-consume-best-practice/Ack-hole.png
asafm, line 1: Any chance to use mermaid for diagrams?

asafm (changes requested):

static/img/blog-consume-best-practice/subscription-types.png
dao-jun, line 1: should be `Shared` not `Shaved`

dao-jun (changes requested):

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+Consumers can subscribe to the topics in four ways (subscription types):
+
+* **Exclusive**
+* **Failover**
+* **Shared**
+* **Key-shared**
+
+The messages are consumed in order for a single partition in Exclusive and Failover modes and out of order for Shared and
+Key-shared mode. The main difference between Exclusive and Failover is that in Exclusive mode, the consumer is exclusive
+to the entire topic, while in Failover mode, the consumer is exclusive to only one partition. Failover mode allows backup
```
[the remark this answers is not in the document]
liangyepianzhou: Yes, it is. For example, there is topic 1 with five partitions; five consumers can subscribe to topic 1 in the failover mode, and only one consumer can subscribe to topic 1 in the exclusive mode. The add consumer logic just like this:
```java
                        case Exclusive:
                            if (dispatcher == null || dispatcher.getType() != SubType.Exclusive) {
                                previousDispatcher = dispatcher;
                                dispatcher = new PersistentDispatcherSingleActiveConsumer(
                                        cursor, SubType.Exclusive, 0, topic, this);
                            }
                            break;
.....
                        case Failover:
                            int partitionIndex = TopicName.getPartitionIndex(topicName);
                            if (partitionIndex < 0) {
                                // For non partition topics, use a negative index so
                                // dispatcher won't sort consumers before picking
                                // an active consumer for the topic.
                                partitionIndex = -1;
                            }

                            if (dispatcher == null || dispatcher.getType() != SubType.Failover) {
                                previousDispatcher = dispatcher;
                                dispatcher = new PersistentDispatcherSingleActiveConsumer(cursor, SubType.Failover,
                                                partitionIndex, topic, this);
                            }
                            break;
```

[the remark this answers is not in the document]
liangyepianzhou: Great! Thanks for the correction.

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+consumer connections that are not consumed. The main difference between Shared and Key-shared is whether their dispatch
```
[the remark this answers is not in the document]
liangyepianzhou: >Key-shared allows to have messages of the same key (e.g. customer-id) sent to the same consumer

Key-shared sent messages with the same key to the same set of consumers instead of the single same consumer.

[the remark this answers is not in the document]
liangyepianzhou: And there are many selector implementations of the selector for the key-shared mode.
For the `ConsistentHashingStickyKeyConsumerSelector`, it could be like the following:
```java
    public Consumer select(int hash) {
        rwLock.readLock().lock();
        try {
            if (hashRing.isEmpty()) {
                return null;
            }

            List<Consumer> consumerList;
            Map.Entry<Integer, List<Consumer>> ceilingEntry = hashRing.ceilingEntry(hash);
            if (ceilingEntry != null) {
                consumerList =  ceilingEntry.getValue();
            } else {
                consumerList = hashRing.firstEntry().getValue();
            }

            return consumerList.get(hash % consumerList.size());
        } finally {
            rwLock.readLock().unlock();
        }
    }
```

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+### Acknowledgment
+
+The messages should be acknowledged after they are fully consumed and processed, and then the messages would not be received
+for the same subscription again. Pulsar provides two ways to acknowledge messages:
+
+* **Cumulative acknowledgment**
+* **Individual acknowledgment**
+
+Cumulative acknowledgment receives a message or a message id as a parameter and marks the messages before the message as
+consumed for this subscription. For multiple-partition topics, the cumulative acknowledgment will work for the single
```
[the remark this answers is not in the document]
liangyepianzhou: The cumulative ack is recommended to be used in the `exclusive` mode and `Failover` mode.   
For the `Exclusive` mode, the messages from all the partitions are sent to the same consumer. But the cumulative ack only marks the messages as processed for the single partition. For the `Exclusive` mode, if they consume 100 messages in a topic with 5 partitions and cumulative ack the last message that is a message in partition4. That does not mean the 100 messages before the last message are processed, it only means the messages in partition 4 before the last message are marked as processed. So I add a notice here.
You can see more information in the following tests.

```java
    @Test
    public void test() throws Exception {
        String subName = "test";
        String topicName = TopicName.get(NAMESPACE1 + "/" + "testCreateTransactionSystemTopic").toString();
        admin.topics().createPartitionedTopic(topicName, 3);
        Producer<byte[]> producer = pulsarClient.newProducer().topic(topicName).create();
        Consumer<byte[]> consumer1 = pulsarClient.newConsumer().receiverQueueSize(100).topic(topicName)
                .subscriptionName(subName)
                .subscriptionType(SubscriptionType.Exclusive).subscribe();

        for (int i = 0; i < 100; i++) {
            producer.newMessage().send();
        }
        Message lastMessage = null;
        for (int i = 0; i < 100; i++) {
            lastMessage = consumer1.receive();
        }
        log.info("ack message {}", lastMessage.getMessageId());
        consumer1.acknowledgeCumulative(lastMessage);

        consumer1.redeliverUnacknowledgedMessages();
        for (int i = 0; i < 30; i++) {
            lastMessage = consumer1.receive();
            log.info("receive message {}", lastMessage.getMessageId());
        }
    }

```

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+partition without impacting other partitions. Individual acknowledgment receives a message or a message id as a parameter
+and only marks this message as consumed for this subscription.
+![img_1.png](../static/img/blog-consume-best-practice/acknowledgement-types.png)
+
+### Messages Redeliver Mechanism
+
+There might be instances where the received messages cannot be processed at this time or some errors happened during processing.
+The client needs to redeliver the unacknowledged messages or a particular message after a delay or immediately.
+Pulsar provides at-least-once semantics when the client does not enable transaction because the client may cache some
```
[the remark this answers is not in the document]
liangyepianzhou: When the transaction is enabled and used, we can ensure all the messages and acknowledgments are persisted to disk. Sorry, I do not know what you mean about the subscription snapshot? Is it related to geo-replication?

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+
+* **ackTimeout**
+* **deadLetterPolicy**
+  * **reconsumeLaterCumulative**
+  * **reconsumeLater**
+* **negativeAcknowledge**
+* **redeliverUnacknowledgedMessages**
+
+The **ackTimeout** is an automatic protection mechanism. If a consumer configured ackTimeout, the messages will be
+auto-redelivered when the received messages are not acknowledged after a long time. It works at the client side, ensuring
```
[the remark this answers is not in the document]
liangyepianzhou: The messages will be dispatched using the same dispatch policy as the last dispatch.
For the shared mode, it will be dispatched to the consumers by round-in.
For the key-shared mode, it will select a set of consumers and send it to one of them. It could be a different consumer and not. The key point is to redeliver unacknowledged messages to the client for re-consumption to avoid ack-hole.

docs/tutorials-redeliver-messages.md
```diff
@@ -0,0 +1,236 @@
...
+Pulsar Consumer API provides four ways to reconsume the unacknowledged messages later:
+
+* **ackTimeout**
+* **deadLetterPolicy**
+  * **reconsumeLaterCumulative**
+  * **reconsumeLater**
+* **negativeAcknowledge**
+* **redeliverUnacknowledgedMessages**
+
+The **ackTimeout** is an automatic protection mechanism. If a consumer configured ackTimeout, the messages will be
```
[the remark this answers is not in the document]
liangyepianzhou: Thanks for the suggestion!
>If not, the client will instruct the broker to redeliver this message to another consumer. 

We can not promise that the message will be delivered to another consumer in the next redelivered, but if they have a set consumer subscribed to the topic, then the messages could be resent to the other consumer after some redelivery.

>QUESTION: Can we instruct max redeliver?

We can not instruct the max redeliver for ack timeout. That was used as an auto-guarantee mechanism.

docs/tutorials-redeliver-messages.md
liangyepianzhou: The acktimeout mechanism works at the client side. It will redeliver unack messages when the connection is active, and the broker will redeliver unack messages when the connection is inactive.

And because the ack timeout mechanism at the client side does not wait for the ack response, that will cause ack-hole while the ack request lost.

docs/tutorials-redeliver-messages.md
liangyepianzhou: I just want to notice the difference between the `reconsumeLaterCumulative ` and `reconsumeLater `.
And notice them about that the `reconsumeLaterCumulative` only resent the copy of the last message to the retry topic.

docs/tutorials-redeliver-messages.md
liangyepianzhou: Yes, this is true. This is one reason that we do not recommend redelivering unack messages when using transactions.

docs/tutorials-redeliver-messages.md
liangyepianzhou: This is related to the choice of subscription type. If users wish to consume a batch of messages in order, and they choose the exclusive mode or failover mode, but they do not send this batch of messages to the same partition, then these messages will be consumed out of order. This is something users do not want to see. To avoid confusing users, it is necessary to explain that before choosing exclusive and failover modes for ordered message consumption, they should ensure that those messages that need to be consumed in order have been sent to the same partition. 

We have not elaborated on how to implement the router. This is still a lead-up to the best practices for consumption.

docs/tutorials-redeliver-messages.md
liangyepianzhou: You're correct, stream processing scenarios indeed often care about the order of messages, and batch processing scenarios also pay attention to the order of messages. However, stream processing scenarios use individual ack to acknowledge individual messages, while batch processing scenarios use cumulative ack to acknowledge messages in bulk. The subscription types Shared and key-shared recommend using individual ack and do not guarantee the order of messages, so we recommend using Shared and key-shared subscription types in stream processing scenarios where the order of messages is not a concern. 

This is not to say that stream processing scenarios do not care about the order of data, but rather that Shared and key-shared types should be used in stream processing scenarios where the order of data is not a concern.

docs/tutorials-redeliver-messages.md
liangyepianzhou: You're correct, we can't solely rely on Pulsar to achieve at least once consumption. However, when we discuss at-least-once consumption in Pulsar, it often refers to the fact that no message will be acknowledged repeatedly. The second ack request for the same message should fail, which is also something that Pulsar transaction needs to ensure.

docs/tutorials-redeliver-messages.md
liangyepianzhou: Ok, I think you are correcting the structure of this article. 
The original article is like the following:
```
# Background knowledge
   ## subscription type
   ## acknowledgment 
   ## redeliver method
# best practice 
  ## shared/ key shared
       ### At-least-once
      ### Exactly-once
 ## Exclusive && Failover
      ### At-least-once
     ### Exactly-once
```
And you want to delete the part of best practice and expand it in the part of the redeliver method, right?

docs/tutorials-redeliver-messages.md
liangyepianzhou: When users use transactions to acknowledge messages, Pulsar must ensure that a message can not be acknowledged twice. The second acknowledgment for one message with a transaction will result in a `TransactionConflictException` and abort the transaction.

docs/tutorials-redeliver-messages.md
liangyepianzhou: The [comment](https://github.com/apache/pulsar-site/pull/875#discussion_r1547880480) you left could be one reason. The other is that one transaction could include many messages, and sending them all to a retry topic or DLQ may not be appropriate when the transaction is aborted.

docs/tutorials-redeliver-messages.md
liangyepianzhou: Maybe the expression here is not very precise. Pulsar has not been able to implement the Exactly-once semantics in `Exclusive` and `Failover` modes, and there have always been repeated acks.

PR #876, opened 2024-04-02
Shuffle testimonials before displaying them

lhotari:
### Motivation

- show a different testimonial each time the page renders

### DIFF
--- a/src/components/pages/HomePage/Users/Users.tsx
+++ b/src/components/pages/HomePage/Users/Users.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import _ from 'lodash'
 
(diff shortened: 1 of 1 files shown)


visortelle:
@lhotari if the parent component will be rendered for some reason, the testimonial list will also be re-shuffled.
If you don't want to think about the parent component behavior, you can use the `useMemo` hook with the `testimonials` dependency.

```tsx
import { useMemo } from 'react'; 
...

const shuffledTestimonials = useMemo(() => _.shuffle(testimonials), [testimonials]);

...

 {Object.values(shuffledTestimonials).flat().map((caseStudy, i) => (
....
```

visortelle (approved):
LGTM

PR #877, opened 2024-04-02
[Doc] Use absolute path when installing functions or using localrun

lhotari:
### Motivation

Absolute paths should be passed to pulsar-admin when creating functions/sinks/sources or when using localrun.
This should be explicitly mentioned in the documentation.

### Additional context

Relative paths with pulsar-admin have never worked because of https://github.com/apache/pulsar/blob/50121e7f7be541f45bb6dc976f51e30658b1cb8d/bin/pulsar-admin#L26


docs/admin-api-functions.md
```diff
@@ -46,7 +46,7 @@ pulsar-admin functions create \
     --inputs test-input-topic \
     --output persistent://public/default/test-output-topic \
     --classname org.apache.pulsar.functions.api.examples.ExclamationFunction \
-    --jar /examples/api-examples.jar
+    --jar $PWD/examples/api-examples.jar
```
david-streamlio, line 49: Is $PWD a safe assumption as to the proper location of the artifact, or should we use a more explicit nomenclature such as `path/to/examples/api-examples.jar` ?

docs/io-cli.md
```diff
@@ -49,8 +49,8 @@ pulsar-admin sources create options
 
 |Flag|Description|
 |----|---|
-| `-a`, `--archive` | The path to the NAR archive for the source. <br /> It also supports url-path (http/https/file [file protocol assumes that file already exists on worker host]) from which worker can download the package.
-| `--classname` | The source's class name if `archive` is file-url-path (file://).
+| `-a`, `--archive` | The absolute path to the NAR archive for the source. <br /> It also supports url-path (http/https/file [file protocol assumes that file already exists on worker host]) from which worker can download the package.
+| `--classname` | The source's class name. Mandatory if `archive` doesn't reference a NAR file.
```
david-streamlio, line 53: I believe the exclusion for "if `archive` doesn't reference a NAR file." only applies if the NAR file has a proper manifest entry, which may or may not be true. 

For now, I would recommend leaving it as "Mandatory" without any exclusions.

docs/io-cli.md
```diff
@@ -81,8 +81,8 @@ pulsar-admin sources update options
...
```
david-streamlio, line 85: Same as above

docs/io-cli.md
```diff
@@ -245,9 +245,9 @@ pulsar-admin sources localrun options
 
 |Flag|Description|
 |----|---|
-| `-a`, `--archive` | The path to the NAR archive for the Source. <br /> It also supports url-path (http/https/file [file protocol assumes that file already exists on worker host]) from which worker can download the package.
+| `-a`, `--archive` | The absolute path to the NAR archive for the Source. <br /> It also supports url-path (http/https/file [file protocol assumes that file already exists on worker host]) from which worker can download the package.
 | `--broker-service-url` | The URL for the Pulsar broker.
-|`--classname`|The source's class name if `archive` is file-url-path (file://).
+|`--classname`|The source's class name. Mandatory if `archive` doesn't reference a NAR file.
```
david-streamlio, line 250: Same

docs/io-cli.md
```diff
@@ -326,9 +326,9 @@ pulsar-admin sinks create options
 
 |Flag|Description|
 |----|---|
-| `-a`, `--archive` | The path to the archive file for the sink. <br /> It also supports url-path (http/https/file [file protocol assumes that file already exists on worker host]) from which worker can download the package.
+| `-a`, `--archive` | The absolute path to the archive file for the sink. <br /> It also supports url-path (http/https/file [file protocol assumes that file already exists on worker host]) from which worker can download the package.
 | `--auto-ack` |  Whether or not the framework will automatically acknowledge messages.
-| `--classname` | The sink's class name if `archive` is file-url-path (file://).
+| `--classname` | The sink's class name. Mandatory if `archive` doesn't reference a NAR file.
```
david-streamlio, line 331: Same

david-streamlio (changes requested):
Looks good overall, just a few minor suggestions regarding the use of $PWD, and the exclusion of classname for NAR files.

docs/admin-api-functions.md
lhotari: I think `$PWD` gives a good hint that it's an absolute path. At least I have found it useful with `pulsar-admin` when I have the files in the current directory and I want to reference the files with an absolute path. I guess there are also other ways.  Perhaps it could be explained to the user that `$PWD` is the current working directory. They will also find out if they try it or google it. :) 
In this particular case, the original `/examples/api-examples.jar` doesn't seem to make much sense. That's why I added the `$PWD` prefix.

docs/io-cli.md
lhotari: > For now, I would recommend leaving it as "Mandatory" without any exclusions.

Ok, I'll just make it "The source's class name". Without any other additions?

lhotari: Done.

docs/io-cli.md
lhotari: Done.

docs/io-cli.md
lhotari: Done.

docs/io-cli.md
lhotari: Done

docs/admin-api-functions.md
lhotari: The `/path/to/file.jar` type of examples have the problem that for the user, it's really a lot of extra steps to find out the absolute path for the file arguments. That's why I prefer `$PWD/file.jar` since it's simple and if someone doesn't already know it, they have the chance to learn it. 

`$PWD` is for a unix shell, but the examples are already assuming a unix shell since the syntax is all unix like.

david-streamlio:
LGTM

PR #878, opened 2024-04-05
[improve]Improve batching message doc

dao-jun:
Improve the batching message doc

Fixes: https://github.com/apache/pulsar/issues/22439

### ✅ Contribution Checklist


dao-jun:
image

docs/concepts-messaging.md
```diff
@@ -427,6 +427,13 @@ Consumer<byte[]> consumer = pulsarClient.newConsumer()
         .subscribe();
 ```
 
+:::note
+
+Send messages by synchronous API `send` will disable batching, and the message will be sent individually.
```
lhotari, line 432: This comment isn't accurate. The batching isn't disabled. The current batch is triggered immediately after sending the message. https://github.com/apache/pulsar/blob/ffff639a1b73a34bbb5115503d4c7783bb2a2770/pulsar-client/src/main/java/org/apache/pulsar/client/impl/TypedMessageBuilderImpl.java#L82-L86

docs/concepts-messaging.md
```diff
@@ -427,6 +427,13 @@ Consumer<byte[]> consumer = pulsarClient.newConsumer()
...
+It is for the purpose of reducing the latency of sending messages and preventing blocking of the producer thread.
```
lhotari, line 433: This is not accurate. "producer thread" is vague in this case. is it an internal thread or what thread is it referring to? I guess a well known concept is "caller's thread" or "calling thread". However, the explanation would have be be better.

One explanation is simply that before the send message returns, the batch would have to be sent and the broker would have to return a message id for the sent message. In most usecases, no more messages could be added to the same batch since the caller thread is blocked so the decision has been made to simply trigger immediate sending of the message when the synchronous API is used. 
I'm not exactly sure how to put this in the docs.

lhotari (changes requested):

lhotari:
> Fixes: https://github.com/apache/pulsar/issues/22439

I don't think that the documentation alone is a sufficient resolution. The Javadoc of the Pulsar Java client needs to be updated too.

docs/concepts-messaging.md
dao-jun: Yes I know it, it's just that my statement is not accurate.

How about change it to: 
```txt
Send messages by synchronous API `send` will trigger the batch to be sent immediately, even the batch is not full.
It is for the purpose of reducing the latency of sending messages and preventing blocking of the caller's thread.
```

lhotari (approved):
Revisited this PR, merging

PR #896, opened 2024-05-10
[feat][doc] PIP-264: Add OpenTelemetry consumer metrics reference

dragosvictor:
### ✅ Contribution Checklist

- [x] I updated the [versioned docs](https://pulsar.apache.org/contribute/document-contribution/#update-versioned-docs) _N/A_


heesung-sn (approved):

docs/reference-metrics-opentelemetry.md
```diff
@@ -402,3 +402,125 @@ The number of pending topic load operations in the broker. When it reaches thres
...
 * Type: UpDownCounter
 * Unit: `{operation}`
+
+### Consumer metrics
+
+#### pulsar.broker.consumer.message.outgoing.count
+The total number of messages dispatched to this consumer.
+* Type: Counter
+* Unit: `{message}`
+* Attributes:
```
asafm, line 412: @dragosvictor Aren't those attributes bigger than existing consumer attributes?
Existing:
```
    private static void writeConsumerMetric(PrometheusMetricStreams stream, String metricName, Number value,
                                            String cluster, String namespace, String topic, String subscription,
                                            Consumer consumer, boolean splitTopicAndPartitionIndexLabel) {
        writeTopicMetric(stream, metricName, value, cluster, namespace, topic, splitTopicAndPartitionIndexLabel,
                "subscription", subscription, "consumer_name", consumer.consumerName(),
                "consumer_id", String.valueOf(consumer.consumerId()));
    }
```

Here:
```
        messageOutCounter = meter
                .counterBuilder(MESSAGE_OUT_COUNTER)
                .setUnit("{message}")
                .setDescription("The total number of messages dispatched to this consumer.")
                .buildObserver();
````
You haven't used the advice to limit the number of attributes exported, so by default all attributed mentioned here will be exported.

The cardinality of those consumer metrics is super high: Client address for example. 
Connected_since should never be used as an attribute, regardless of it being super high cardinality.

asafm (changes requested):

docs/reference-metrics-opentelemetry.md
dragosvictor: Resolved as per https://github.com/apache/pulsar/pull/22837, thanks for pointing it out!

heesung-sn:
@asafm feel free to raise a concern if you still have questions.

PR #939, opened 2024-07-10
add topic naming note

yuweisung:
### ✅ Contribution Checklist

### DIFF
--- a/docs/admin-api-topics.md
+++ b/docs/admin-api-topics.md
@@ -36,7 +36,7 @@ non-persistent://tenant/namespace/topic
 
 :::note
 
-Topic naming: due to backward compability, some special characters, eg., "/", is allowed as a part of topic name. But we recommended that you should avoid using special characters as a part of topic name.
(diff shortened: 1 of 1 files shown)


docs/admin-api-topics.md
```diff
@@ -34,6 +34,10 @@ Non-persistent topics are used in applications that only consume real-time publi
 non-persistent://tenant/namespace/topic
 ```
 
+:::note
+
+Topic naming: due to backward compability, some special characters, eg., "/", is allowed as a part of topic name. But we recommended that you should avoid using special characters as a part of topic name.
```
tisonkun, line 39:
```suggestion
Topic naming: due to backward compability, some special characters, eg., "/", is allowed as a part of topic name. But we recommended that you should avoid using special characters as a part of topic name.

:::
```

docs/admin-api-topics.md
```diff
@@ -34,6 +34,12 @@ Non-persistent topics are used in applications that only consume real-time publi
...
```
tisonkun, line 39:
```suggestion
Topic naming: due to backward compatibility, some special characters, e.g., "/," are allowed as a part of the topic name. But it's recommended not to use special characters as a part of the topic name.
```

shibd (approved):

PR #954, opened 2024-08-25
[fix][doc] Remove not needed article in Key_Shared sticky mapping documentation

pdolif:
### ✅ Contribution Checklist


lhotari:
Thanks for the improvement! 

btw. The website has now been migrated to Docusaurus 3 and MDX 3 in #967 . MDX 3 is more strict about escaping of expressions that could be mixed up as code expressions. For example in `<=`, it's necessary to use html entities to turn `<` into `&lt;`. ([more details](https://docusaurus.io/docs/migration/v3#common-mdx-problems))

We will actively maintain docs for 4.0.x, 3.3.x (until December), 3.0.x and next (== main), so there isn't need to make updates for older versions for minor issues. I had to update all content for the Docusaurus 3 migration since the site compilation would crash due to invalid content issues which passed the previous Docusaurus 2 parser.

PR #866, opened 2024-03-20
Use correct REST API URLs in docs

visortelle:
### ✅ Contribution Checklist

- [x] I read the [contribution guide](https://pulsar.apache.org/contribute/document-contribution/)
- [ ] I updated the [versioned docs](https://pulsar.apache.org/contribute/document-contribution/#update-versioned-docs)

---

- [x] Display proper URL in docs

  Screenshot 2024-03-20 at 9 33 53 PM
  Screenshot 2024-03-20 at 9 34 07 PM

- [ ] Add class name prefix to `<189 entries> * <supported version count>`

  Screenshot 2024-03-20 at 9 31 45 PM

https://github.com/apache/pulsar/issues/21573
https://github.com/apache/pulsar/pull/19193
https://github.com/apache/pulsar/issues/22289

PR #970, opened 2024-10-28
Fix REST API urls in documentation

lhotari:
Fixes apache/pulsar#21573

All REST API urls are incorrect in the documentation. This PR fixes the URLs for next, 4.0.x and 3.0.x.

This PR adds a custom Remark plugin that handles logic to lookup the correct URL from the Swagger JSON file so that it's always correct. The plugin is written in Typescript and to do that, the docusaurus config was converted to `.ts` extension and made compatible.

There's a remaining problem that the Swagger files contain a lot of duplicate operations even after apache/pulsar#19193. The links to Swagger documentation won't work when there are duplicates, which is very common. A replacement will be needed in apache/pulsar for the Swagger generation to fix it. The duplicate operation names required adding ways to find the correct operation in the custom Remark plugin.

This PR replaces the previous draft PR #866.

### DIFF
--- a/docs/admin-api-brokers.md
+++ b/docs/admin-api-brokers.md
@@ -45,7 +45,7 @@ localhost:8080
 </TabItem>
 <TabItem value="REST API">
 
-{@inject: endpoint|GET|/admin/v2/brokers/:cluster|operation/getActiveBrokers?version=@pulsar:version_number@}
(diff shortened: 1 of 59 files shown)


lhotari:
some remaining work to fix the issue at https://github.com/apache/pulsar-site/blob/8df377998baa66ea709bbc6ea26a4aa627f9bf5d/src/utils/index.js#L8 . I'll fix that.

PR #973, opened 2024-11-06
Fix retention and backlog quota diagram

shibd:
The current retention and backlog quota diagram does not match the actual implementation.

Refer to issue: https://github.com/apache/pulsar/issues/22473#issuecomment-2426787328
Discuss thread: https://lists.apache.org/thread/g3hc3o0kmn1yvcfdj6xnlbq4vsdskx9o

### Preview

Retention policy
image

Backlog quotas
image


docs/cookbooks-retention-expiry.md
```diff
@@ -205,7 +205,7 @@ The following retention policies are available:
 | `consumer_backlog_eviction` | The broker will begin discarding backlog messages                                                 |
 
 > #### Beware the distinction between retention policy types
-> As you may have noticed, there are two definitions of the term "retention policy" in Pulsar, one that applies to persistent storage of messages not in backlogs, and one that applies to messages within backlogs.
+> As you may have noticed, the scope of retention includes the backlog, so retention must be greater than the backlog quota; otherwise, Pulsar will throw an error when set both.
```
lhotari, line 208: Fix grammar in this sentence. "when set both" is not correct. "As you may have noticed" could be omitted. Suggestion:
```
:::note
The scope of retention includes the backlog, so retention must be greater than the backlog quota. Otherwise, Pulsar will throw an error when both are set.
:::
```

docs/cookbooks-retention-expiry.md
```diff
@@ -37,7 +37,7 @@ Pulsar's [admin interface](admin-api-overview.md) enables you to manage both ret
 By default, when a Pulsar message arrives at a broker, the message is stored until it has been acknowledged on all subscriptions, at which point it is marked for deletion. You can override this behavior and retain messages that have already been acknowledged on all subscriptions by setting a *retention policy* for all topics in a given namespace. Retention is based on both a *size limit* and a *time limit*.
 
 The diagram below illustrates the concept of message retention.
-![Concept of message retention in Pulsar.](/assets/retention.svg)
+![Concept of message retention in Pulsar.](/assets/retention-new.png)
```
lhotari, line 40: why are you switching to use .png files? How are these files created? What is the original source of the diagram if someone else would like to edit the diagram?

lhotari: Where are the sources for these diagrams?

lhotari: How did you regenerate the diagram?

shibd: I recreated it locally using OmniGraffle, and it supports generating SVG format.

lhotari: Would it be possible to share the original OmniGraffle files? It would be useful if anyone could be able to maintain the diagrams that we have in the docs.

poorbarcode (approved):

dao-jun (approved):

docs/cookbooks-retention-expiry.md
shibd: Of course, but should we upload it directly to this repository? Each file is about 3MB, and I'm worried it might get too large in the future. Can we put it in SVN instead? For example: https://dist.apache.org/repos/dist/release/pulsar/pulsar-site/orignal-diagram

lhotari: I don't think that 3MB is much in repository size. How many files are there for the diagram files?
Let's keep them in pulsar-site git repository, for example in `diagrams` sub directory. That would get ignored in the Docusaurus build so the website size wouldn't increase.

docs/cookbooks-retention-expiry.md
```diff
@@ -37,7 +37,7 @@ Pulsar's [admin interface](admin-api-overview.md) enables you to manage both ret
 By default, when a Pulsar message arrives at a broker, the message is stored until it has been acknowledged on all subscriptions, at which point it is marked for deletion. You can override this behavior and retain messages that have already been acknowledged on all subscriptions by setting a *retention policy* for all topics in a given namespace. Retention is based on both a *size limit* and a *time limit*.
 
 The diagram below illustrates the concept of message retention.
-![Concept of message retention in Pulsar.](/assets/retention.svg)
+![Concept of message retention in Pulsar.](/assets/retention-new.svg)
```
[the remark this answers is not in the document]
shibd: Since I haven't modified the references for older versions, like 2.10, I've kept the original files.

docs/cookbooks-retention-expiry.md
shibd: that makes sense. I push to ./diagrams directory.

docs/cookbooks-retention-expiry.md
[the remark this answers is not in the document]
shibd: @lhotari WHYT? 

Do we need to update all versions of the documentation to point to the new diagrams? This way, I can remove the original diagrams.

PR #978, opened 2024-12-17
Avoid link to apache archive server directly

Shawyeok:
Fixed https://github.com/apache/pulsar/issues/23710, following [the guide][1] use `closer.lua` link in download instructions.

[1]: https://infra.apache.org/release-download-pages.html#links

### DIFF
--- a/docs/administration-pulsar-shell.md
+++ b/docs/administration-pulsar-shell.md
@@ -21,10 +21,10 @@ It's great for quickly switching between different clusters, and can modify clus
 
 
 ## Install Pulsar Shell
-Download the tarball from the [download page](pathname:///download) and extract it.
+Download the tarball from the [download page](pathname:///download#shell) and extract it.
 
 ```shell
-wget https://archive.apache.org/dist/pulsar/pulsar-@pulsar:version@/apache-pulsar-shell-@pulsar:version@-bin.tar.gz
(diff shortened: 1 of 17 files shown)


lhotari (changes requested):
Please make the change also to the maintained versions in `versioned-docs/4.0.x` and `versioned-docs/3.0.x` directories. 
More details in https://github.com/apache/pulsar-site?tab=readme-ov-file#updating-versioned-docs and https://pulsar.apache.org/contribute/document-contribution/#update-versioned-docs .

The recommended approach is to first update the docs directory document and then use the tool to apply the patch to maintained versions.

Shawyeok:
> Btw. I'm not sure if downloads.apache.org is an acceptable link. In https://github.com/apache/pulsar/issues/23710, there's nothing that says that the link should be downloads.apache.org link. Instead [in the guide](https://infra.apache.org/release-download-pages.html#links), it says that closer.lua must be used. (checksums and signatures should use the downloads.apache.org URL).

@lhotari 
Sorry for the late reply. It seems you edited my last comment, so I didn’t get any notifications about the edit.

I’ve updated the download instructions by adding a link with `closer.lua` and removing the `wget` command. This allows users to download the Pulsar release with just one or two clicks, making it more convenient.

lhotari (approved):

PR #979, opened 2024-12-17
[fix] Fix init cluster command

nodece:
### ✅ Contribution Checklist

### Motivation

In Pulsar 4.0, we switch the commander parse from the jcommander to the picocli, which is more strict than the jcommander. When the docker-compose.yml, if we are using the command line continuation feature, we must remove the extra spaces and line breaks. I reported this issue in the picocli a long time ago, the team thinks this is a docker-compose issue, please see https://github.com/remkop/picocli/issues/2348.

Right now, we can use `bash -c` to avoid this issue.


lhotari (changes requested):
Please make the change also to the `docs/` directory. 
More details in https://github.com/apache/pulsar-site?tab=readme-ov-file#updating-versioned-docs and https://pulsar.apache.org/contribute/document-contribution/#update-versioned-docs .

The recommended approach is to first update the docs directory document and then use the tool to apply the patch to maintained versions.

nodece:
@lhotari Done.

lhotari (approved):

PR #982, opened 2024-12-24
Fix file source connector example's keepFile property value

CaveEcho:
if ["keepFile": true] starts The connector "The property keepFile must be false if the property processedFileSuffix is set"

### ✅ Contribution Checklist


liwenjielala:
Description
While following the YAML configuration example provided in the Pulsar documentation, I encountered a situation where setting keepFile to true while also setting processedFileSuffix causes an error:
“The property keepFile must be false if the property processedFileSuffix is set.”

This situation arises because the keepFile property must be set to false if processedFileSuffix is configured. To resolve this, I would like to suggest two possible approaches:

1. Set keepFile to false when processedFileSuffix is configured, in order to avoid the error.
2. Alternatively, if it’s not strictly necessary, remove the processedFileSuffix setting from the YAML example to help simplify the configuration.
I hope these suggestions are helpful and clarify the intended configuration. Thank you for considering them!

dave2wave:
@lhotari ready to merge?

lhotari (approved):

PR #989, opened 2025-02-18
Update WEChat description

dave2wave:
I think that the WeChat channel is not an official channel. We can discuss amongst the PMC if there is not agreement


lhotari (approved):

lhotari:
@tisonkun Please review

tisonkun (approved):

tisonkun:
It's actually "Offical" in "WeChat Official Accounts" but since it generates confusion I agree to change the content since the one after doesn't hurt.

PR #1117, opened 2026-04-16
Docs: warn that negativeAcknowledge redelivery counter resets on dispatcher close

### LINKED ISSUE #25533: Docs: warn that DeadLetterPolicy.maxRedeliverCount is silently unreliable with negativeAcknowledge under bundle unload / dispatcher close
## Summary

The docs for `negativeAcknowledge` and `DeadLetterPolicy.maxRedeliverCount` don't warn that the redelivery counter is kept exclusively in client and broker RAM, and is lost whenever the subscription's dispatcher is closed. In any cluster with `loadBalancerEnabled=true` (the default), bundles are routinely unloaded and rebalanced, so `maxRedeliverCount` acts as a soft guideline at best and the DLQ is in practice never reached for long-lived poison messages.

This is **not** a behavior bug: `reconsumeLater(...)` with `enableRetry(true)` works correctly and persists its counter as a `RECONSUMETIMES` property on the retry topic, in BookKeeper. The problem is that nothing in the documentation tells a reader to prefer `reconsumeLater` over `negativeAcknowledge` when they care about DLQ guarantees — the two are presented as interchangeable retry mechanisms.

Related discussion: #23944 covers a different angle of the same broader "redelivery count is unreliable" theme (consumer crashes vs. dispatcher close), and proposes a behavior change. This issue is intentionally narrower and **docs-only** — even if #23944 lands a behavior fix, the docs should warn current users today.

## Why this matters (real-world)

Our team ran a consumer with `negativeAcknowledge` + `DeadLetterPolicy(maxRedeliverCount=10)` in production for months before realizing the DLQ was never reached. The subscription accumulated ~400k messages in the backlog because every bundle unload reset the in-memory redelivery counter, and the backoff cycle was longer than the typical time between unloads. We only understood the mechanism after writing an instrumented reproduction against a standalone broker. **A single paragraph in the docs would have saved us that time**, which is why I'm opening this.

## Reproduction

Self-contained, runnable: https://github.com/ng-galien/bug-pulsar

| # | Client | Broker | Event applied | DLQ reached? |
|---|---|---|---|---|
| A | 2.11.0 | 2.11.0 | \`docker-compose restart pulsar\` | ❌ 0 msgs |
| D | 2.11.0 | 2.11.0 | \`pulsar-admin topics unload\` (broker stays up) | ❌ 0 msgs |
| V3 | 2.11.0 | 3.3.9 | \`pulsar-admin topics unload\` | ❌ 0 msgs |
| V3 (\`-Pclient-v3\`) | 3.3.9 | 3.3.9 | \`pulsar-admin topics unload\` | ❌ 0 msgs |

After the unload, the same \`MessageId\`s come back with \`redeliveryCount=0\` because the \`InMemoryRedeliveryTracker\` attached to the dispatcher is destroyed and recreated empty. Scenario B in the same repo validates the counterpart: \`reconsumeLater(...)\` + \`enableRetry(true)\` is unaffected by either a broker restart or a topic unload, because the counter lives on the retry topic as a persisted message property.

The bug reproduces identically on the latest 3.3.x line, with both the 2.11 and 3.3 clients.

## What I'm asking for

Docs-only. No behavior change, no API change.

1. In the *Negative acknowledgement* page, add a warning callout stating:
   - the redelivery counter is kept only in client (\`NegativeAcksTracker\`) + broker (\`InMemoryRedeliveryTracker\`) RAM;
   - any event that closes the subscription dispatcher (broker restart, bundle unload, bundle rebalance, last consumer disconnect) resets the counter;
   - therefore \`DeadLetterPolicy.maxRedeliverCount\` cannot be relied on as a hard upper bound when \`negativeAcknowledge\` is the retry mechanism;
   - recommend \`reconsumeLater(...)\` with \`enableRetry(true)\` when the DLQ must actually be reached.

2. In the javadoc for \`Consumer.negativeAcknowledge(...)\` and \`DeadLetterPolicy.Builder.maxRedeliverCount(...)\`, add a one-liner that links to that callout.

3. Optional: update the *Retry letter topic* page to cross-reference the nack page and explain *why* one would choose one over the other — the current docs describe them as alternatives without explaining the persistence difference, which is arguably the entire point.

Happy to submit a docs PR with proposed wording if a maintainer agrees this is worth documenting.

## Versions covered by the repro

- Broker: \`apachepulsar/pulsar:2.11.0\` and \`apachepulsar/pulsar:3.3.9\` (both reproduce)
- Client: \`pulsar-client:2.11.0\` and \`pulsar-client:3.3.9\` (both reproduce)
- Java: 21 (Temurin), macOS + Docker

ng-galien:
### ✅ Contribution Checklist

## Summary

Add caution admonitions to warn users that `DeadLetterPolicy.maxRedeliverCount` is silently unreliable when `negativeAcknowledge` is used as the retry mechanism.

The redelivery counter is kept only in memory (client-side and broker-side). It resets to zero on broker restart, bundle unload/rebalance, topic unload, or consumer disconnect. In practice, failing messages can be redelivered indefinitely and the dead letter topic may never be reached.

Three sections updated in `concepts-messaging.md`:

1. **Negative acknowledgment** — new `:::caution` block explaining the issue and recommending `reconsumeLater` with `enableRetry(true)`
2. **Retry letter topic** — expanded existing note to clarify that retry letter topic is the only mechanism that guarantees `maxRedeliverCount` is honored
3. **Dead letter topic** — new `:::caution` block warning that `maxRedeliverCount` may not be honored without `enableRetry(true)`

Changes propagated to versioned docs (3.0.x, 4.0.x, 4.2.x) via `docs-tool.sh`.

Verified with a self-contained reproduction across Pulsar 2.11.0, 3.3.9, and 4.2.0 (broker and client): https://github.com/ng-galien/bug-pulsar

Closes apache/pulsar#25533

### DIFF
--- a/docs/concepts-messaging.md
+++ b/docs/concepts-messaging.md
@@ -156,6 +156,16 @@ If batching is enabled, all messages in one batch are redelivered to the consume
 
 :::
 
+:::caution Redelivery counter is not persisted
+
+With `negativeAcknowledge`, the redelivery counter is kept **only in memory**. It resets to zero whenever the broker closes the subscription dispatcher — which happens on broker restart, bundle unload/rebalance, topic unload, or consumer disconnect.
+
+As a result, `DeadLetterPolicy.maxRedeliverCount` **may never be reached** and failing messages can be redelivered indefinitely without reaching the [dead letter topic](#dead-letter-topic).
+
+If your application needs a reliable retry limit, use [`reconsumeLater`](#retry-letter-topic) with `enableRetry(true)` instead — the retry count is persisted as a message property and survives all of the above events.
+
+:::
+
 ### Acknowledgment timeout
 
 :::note
@@ -315,7 +325,7 @@ consumer.reconsumeLater(msg, customProperties, 3, TimeUnit.SECONDS);
 :::note
 
 *  Currently, retry letter topic is enabled in Shared subscription types.
-*  Compared with negative acknowledgment, retry letter topic is more suitable for messages that require a large number of retries with a configurable retry interval. Because messages in the retry letter topic are persisted to BookKeeper, while messages that need to be retried due to negative acknowledgment are cached on the client side.
+*  Compared with negative acknowledgment, retry letter topic is more suitable for messages that require a reliable retry limit. Messages in the retry letter topic have their retry count persisted as a message property, while the redelivery counter used by `negativeAcknowledge` is kept only in memory and [can be reset unexpectedly](#negative-acknowledgment). This makes `reconsumeLater` with `enableRetry(true)` the only mechanism that guarantees `maxRedeliverCount` is honored and the [dead letter topic](#dead-letter-topic) is eventually reached.
 
 :::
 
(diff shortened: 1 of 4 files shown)


lhotari (approved):
LGTM, thanks for the contribution @ng-galien

PR #1123, opened 2026-04-22
fix(downloads): point client release tables at the tarball instead of the directory

### LINKED ISSUE #23836: [Doc] https://pulsar.apache.org/download/ incorrect links for Pulsar C++ Client
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

The links under Pulsar C++ Client all point to directories, rather than individual files.

### What is your suggestion?

The links should point to the files as per the asc and sha512 entries

### Any reference?

_No response_

- [ ] I'm willing to submit a PR!

singhvishalkr:
﻿## Motivation

Fixes apache/pulsar#23836.

Looking at [`src/components/downloads.tsx`](https://github.com/apache/pulsar-site/blob/main/src/components/downloads.tsx), the four client release tables &mdash; C++, Go, Node, Python &mdash; set `link: url`, where `url` is the Apache archive **directory** URL (`.../pulsar-client-cpp-X.Y.Z/`). The sibling `asc` and `sha512` fields correctly use `tarPath` (the tarball file).

The result on [pulsar.apache.org/download](https://pulsar.apache.org/download/#old-releases) is a row where:

| Column      | Resolves to                        |
|-------------|------------------------------------|
| Download    | directory listing (a UI of files)  |
| asc         | `.tar.gz.asc`                      |
| sha512      | `.tar.gz.sha512`                   |

Which is exactly the mismatch the issue reports: *"The links under Pulsar C++ Client all point to directories, rather than individual files... The links should point to the files as per the asc and sha512 entries."*

The same bug is present in the Go, Node, and Python tables &mdash; same line pattern, same function body. Fixing all four in one patch keeps the tables consistent.

## Modifications

Change `link: url` &rarr; `link: tarPath` in:

- `CppReleasesDownloadTable` (line 185 before patch)
- `GoReleasesDownloadTable`  (line 204)
- `NodeReleasesDownloadTable` (line 225)
- `PythonReleasesDownloadTable` (line 245)

## Verifying this change

Open `/download` in a built docs site &mdash; hovering/clicking the release name in the four client tables should now navigate to the `apache-pulsar-client-<lang>-<version>.tar.gz` file instead of the enclosing directory. `.asc` and `.sha512` columns are unchanged.

## Does this pull request potentially affect one of the following parts:

- [x] `The docs`
- [ ] Anything else

## Documentation

- [x] `doc` &mdash; rendered download page.


lhotari (changes requested):
This requires some more changes. The table should be similar to "Older releases" above the tables where there would be links for the binary (asc, sha512), source (asc, sha512) and release notes.

singhvishalkr:
Thanks for the review @lhotari! Before I push the reshape I wanted to flag what the dist layout actually exposes for each client, so the final table matches reality:

| Client | Binary archive at `archive.apache.org/dist` | Source archive | Versioned release notes under `release-notes/versioned/` |
|---|---|---|---|
| C++ | none (no RPM/DEB in the tree; some versions ship `apache-pulsar-client-cpp-${v}.tar.gz` only) | `apache-pulsar-client-cpp-${v}.tar.gz` | **yes** (`client-cpp-*.md` exists) |
| Go | none | `apache-pulsar-client-go-${v}-src.tar.gz` | no (release notes live on GitHub releases) |
| Node | `napi-*` prebuilts per platform (darwin/linux/win, arm64/x64, glibc/musl) plus a source tarball | `apache-pulsar-client-node-${v}.tar.gz` | no (GitHub releases) |
| Python | none on dist (wheels ship on PyPI, not ASF) | `pulsar-client-python-${v}.tar.gz` | **yes** (`client-python-*.md` exists) |

Given that, two shapes make sense -- happy to ship whichever you prefer:

**Option A** -- reuse `OldReleaseTable` verbatim with 4 columns: `Release | Binary | Source | Release Notes`.
- C++ + Python: leave Binary empty, point Source at the tarball, link Release Notes to the versioned page.
- Go: leave Binary empty, point Source at the `-src.tar.gz`, point Release Notes at `https://github.com/apache/pulsar-client-go/releases/tag/v${v}`.
- Node: point Binary at the source tarball *and* summarise napi prebuilts in a footnote (or a nested disclosure link), Source at the same tarball, Release Notes at `https://github.com/apache/pulsar-client-node/releases/tag/v${v}`. Node is the only client where "Binary" has a real meaning on dist, but it's N platform-specific rows, not one.

**Option B** -- extend the current `ReleaseTable` to also carry a `releaseNote` column; keep a single "Download" column because 3 of the 4 clients are source-only anyway, so a "Binary" column would be empty for them.

My lean is **Option A** -- it keeps the page visually consistent with "Older releases" above, which is what you asked for -- with Node handled by linking to `napi-*` listings in a sub-line. Want me to go with A and submit, or do you prefer the B path?

Either way, would you like me to keep the existing one-line bug fix (tarball instead of directory) in this PR or split it out as a smaller prior PR so the reshape lands cleanly?

singhvishalkr:
Reshaped the client tables to match the Older releases layout -- @lhotari per your review. Two files changed on this push:

- `src/components/OldReleaseTable.js`: the Binary column is now rendered only when at least one row has a `binary` field set. The existing callers for Pulsar server (`ArchivedPulsarDownloadTable`) and Pulsar Manager (`ArchivedPulsarManagerDownloadTable`) keep their 4-column layout (Release / Binary / Source / Release notes) because every row there has a binary tarball. When all rows are source-only, the component drops the Binary column and renders 3 columns instead.
- `src/components/downloads.tsx`: `Cpp`/`Go`/`Node`/`Python` `ReleasesDownloadTable` now emit the `OldReleaseTable` shape (`source`, `sourceText`, `sourceAsc`, `sourceSha`, `sourceShaText`, `releaseNote`) instead of the flat `ReleaseTable` shape. For release notes I point at each client repo's GitHub releases tag (`https://github.com/apache/pulsar-client-<lang>/releases/tag/v<version>`) since that's the authoritative changelog for these projects (pulsar-site only has `release-notes/versioned/client-cpp-*.md` and `client-python-*.md` for some versions, nothing for `client-go`/`client-node`, so one consistent source keeps every row clickable).

Note on the honest gap: Apache dist publishes source tarballs for all four clients and nothing else, so the Binary column would be empty for every row. I went with hiding the column rather than rendering empty cells; happy to switch to rendering a dash/placeholder if you'd rather keep the 4-column header for visual consistency with the Pulsar table above it.

Also kept the original one-line fix (`link: url` -> `link: tarPath`) folded into the new data shape so the table entries still point at the actual `.tar.gz` rather than the directory.

lhotari:
There's an error. Please test your changes locally by previewing the site with `./preview.sh`.
```
Compiled with problems:
×
ERROR in ./src/components/downloads.tsx
  × Module build failed (from builtin:swc-loader):
  ├─▶   ×   × Unexpected character '\0'
  │     │    ╭─[[36;1;4m/Users/lari/workspace-pulsar/pulsar-site/src/components/downloads.tsx:1:1]
  │     │  1 │ import React from 'react'
  │     │    · [35;1m ▲
  │     │  2 │ import pulsarReleases from '@site/releases.json'
  │     │  3 │ import connectors from '@site/data/connectors'
  │     │  4 │ import cppReleases from '@site/data/release-cpp'
  │     │    ╰────
  │     │   × Expected ';', '}' or <eof>
  │     │    ╭─[[36;1;4m/Users/lari/workspace-pulsar/pulsar-site/src/components/downloads.tsx:1:1]
  │     │  1 │ import React from 'react'
  │     │    · [35;1m┬[33;1m─
  │     │    · [35;1m╰── [35;1mThis is the expression part of an expression statement
  │     │  2 │ import pulsarReleases from '@site/releases.json'
  │     │  3 │ import connectors from '@site/data/connectors'
  │     │  4 │ import cppReleases from '@site/data/release-cpp'
  │     │    ╰────
  │     │
  │   
  ╰─▶   × Syntax Error
```

singhvishalkr:
Pushed `bfc35b8` — my earlier reshape commit `c15b068` had written `src/components/downloads.tsx` as UTF-16LE instead of UTF-8, which is why the SWC loader tripped on `Unexpected character '\0'` at position 1:1 (the `i` of `import` followed by a null byte was the interleaved UTF-16 high byte). The `OldReleaseTable.js` change in the same commit was fine because it wasn't re-encoded. Re-encoded the file back to UTF-8 without a BOM and force-pushed; `./preview.sh` should build now. Sorry for the noise.

Content diff vs `main` is unchanged — same 4 client tables switched to `OldReleaseTable` shape, same `source`/`sourceText`/`sourceAsc`/`sourceSha`/`releaseNote` fields, same Binary-column-hidden behaviour in `OldReleaseTable.js`.

lhotari (changes requested):
After all, this doesn't make sense at least for CPP client. 
For example, in https://archive.apache.org/dist/pulsar/pulsar-client-cpp-4.1.0/ there are also pre-built binaries available. It would be a larger change to handle links to all possible binaries and having that in the table wouldn't be usable. Something where the table row gets expanded and contains the links could work.

singhvishalkr:
You were right on the CPP point -- trying to shoehorn every prebuilt into a flat table was the wrong direction. Pushed `9c3317be` that takes the row-expansion route you suggested.

Changes on this push:

- Reverted `src/components/OldReleaseTable.js` back to its pre-PR shape (this PR no longer needs to touch the Pulsar-server / Pulsar-Manager tables at all).
- Added `src/components/ClientReleasesTable.tsx`. Each client release row is now:
  - Collapsed: `Release | source tarball (asc, sha512) | Release notes`, with an expand toggle in a leading cell.
  - Expanded: shows the Apache archive directory URL for that version (`archive.apache.org/dist/pulsar/pulsar-client-<lang>-${version}/`), which is where every prebuilt lives, plus a short hint:
    - CPP: "Pre-built DEB, RPM and APK packages (where published) live alongside the source tarball in the directory above."
    - Node: "Platform-specific napi prebuilts (darwin/linux/win, arm64/x64, glibc/musl) are published alongside the source tarball in the directory above."
    - Go and Python: no hint (source-only on dist).
- `CppReleasesDownloadTable`, `GoReleasesDownloadTable`, `NodeReleasesDownloadTable`, `PythonReleasesDownloadTable` in `downloads.tsx` now render through `ClientReleasesTable`. The original narrow bug (row link pointing at the directory rather than the tarball) stays fixed because the collapsed-row tarball link uses `tarPath` and the directory is only reached via the explicit expand toggle.

Implementation: MUI `Collapse` + `IconButton` with the `KeyboardArrowDown`/`Up` icons (both already listed in `package.json` under `@mui/icons-material` and `@mui/material`). Each row keeps its own `useState` so rows expand independently.

Local verification on this push (addressing the previous encoding footgun on `c15b068`):

- `tsc --noEmit` across the project -- no new errors on `ClientReleasesTable.tsx` or `downloads.tsx` (pre-existing errors in `src/theme/Blog*/**` are unchanged, unrelated to this PR).
- SWC parse (`@swc/core` `transformSync` with `syntax: 'typescript', tsx: true`) -- both files parse cleanly. This is the same loader that flagged `c15b068` at position 1:1; running it locally catches the regression class that blew up `c15b068` before I push.
- Blob byte check after push: first bytes of both files on the GitHub contents API are `69 6D 70 6F 72 74` / no BOM / no UTF-16 markers.

If you'd rather have the collapsed state render the prebuilt hint inline (e.g. next to the release notes column instead of gated behind expand), or want the CPP table to list the per-distro binary paths inline once the naming convention is stable enough to hardcode, I can iterate in a follow-up push.

lhotari:
The visual appearance is different from the other release tables:
image

- the row height is larger than it is in the other tables
- even and odd rows should use different background color
- it would be logical to have similar columns as the other tables in the same order
  - add a column title "Binaries".
- A simpler approach could be to just have a link to the directory instead of having the dropdown. The link title could be the short name of the directory, for example [pulsar-client-cpp-4.1.0/](https://archive.apache.org/dist/pulsar/pulsar-client-cpp-4.1.0/). This would mean that we could go back to a similar table as it was originally and just add the source column.

singhvishalkr:
Reshaped the client tables per your screenshot notes — pushed `5b59795a` as a single clean commit off `main` (the three earlier commits are gone, so the diff is now `ClientReleasesTable.tsx` + `downloads.tsx` only).

- Columns reduced to `Release | Source | Release notes`, matching the three-column shape of the `ReleaseTable` used for the Pulsar server / offloaders / connectors tables on the same page. The expand chevron moved inside the Release cell so the table doesn't carry a fourth, header-less column.
- Odd rows now tint at `rgba(0, 0, 0, 0.03)`; the collapsible sibling row inherits the same tint, so an expanded entry visually groups with its trigger.
- Row height: `paddingY: 0.5` on the three cells plus a smaller `IconButton` (`size="small"`, `padding: 0.25`, `fontSize="inherit"`) so a collapsed client row now matches the height of an `OldReleaseTable` row rather than the default ~48px the chevron was forcing.

Row expansion behaviour is unchanged — click the chevron, get the Apache archive directory URL plus (for CPP/Node) the pre-built binaries hint. The tarball link in the `Source` column still bypasses the directory listing, which is the original issue this PR set out to fix.

lhotari:
@singhvishalkr I'm sorry I missed the issue before. There's a comment in https://github.com/apache/pulsar/issues/23836#issuecomment-4395031199

singhvishalkr:
Saw your note on apache/pulsar#23836. This PR is still the active fix for the download-page links; the branch head is 5b59795a. I re-ran `./preview.sh` on that commit before replying here.

Annaxs (commented):
Hi @singhvishalkr — reviewed the PR. Here's what I found:

**Core fix ✅**
The `link: url → link: tarPath` change is correct across all four tables. I verified the tarball filenames against the actual Apache archive:
- CPP: `apache-pulsar-client-cpp-{version}.tar.gz` ✅
- Go: `apache-pulsar-client-go-{version}-src.tar.gz` ✅
- Node: `apache-pulsar-client-node-{version}.tar.gz` ✅
- Python: `pulsar-client-python-{version}.tar.gz` ✅

**New `ClientReleasesTable` component ✅**
The expandable row design is a smart call — it fixes the main download link while still surfacing the directory URL for users who need pre-built binaries (DEB/RPM for CPP, napi packages for Node). The `binariesHint` on CPP and Go rows is a nice touch.

TypeScript compiles cleanly for both new files — no new errors introduced.

**One thing worth checking**
The Node archive doesn't appear to have source tarballs for some older versions (e.g. `pulsar-client-node-1.9.0` has no directory in the archive). If a version's `tarPath` resolves to a 404, the download link would silently fail. It may be worth checking whether the older Node entries in `releases.json` all have a corresponding source tarball, or adding a note in the `binariesHint` for versions where only napi prebuilts are available.

Overall the PR looks good to me. Happy to help test further or address the edge case above if useful.

singhvishalkr:
@Annaxs thanks for the thorough pass on the four tables and the tarball naming check against archive.apache.org.

I pulled upstream `main` into this branch (merge commit `8fa6e4e1a`) so CI tracks the current site, then pushed `1820f461f`: a short comment above the Node `excludedVersions` list explaining that those tags predate the current per-version directory layout on dist, plus one extra sentence on the Node binaries hint pointing people at the expanded directory row if a hot-linked tarball 404s.

On the 1.9.0 concern: I ran HEAD requests against `apache-pulsar-client-node-1.9.0.tar.gz`, `1.8.2`, and `1.17.0` on archive.apache.org; all returned 200. If you still see a mirror edge case, paste the version and we can add it to the exclusion list or split the row logic.

@lhotari this branch is rebased on current `main`; ready whenever you have review bandwidth.

Annaxs (approved):
Tested locally on commit `1820f461` via `./preview.sh` and verified the following:

**Download page renders without errors ✅**
The dev server starts cleanly and `http://localhost:3000/download` loads correctly.

**Source tarball links are correct across all four tables ✅**
- CPP: `apache-pulsar-client-cpp-{version}.tar.gz` — points to file, not directory
- Go: `apache-pulsar-client-go-{version}-src.tar.gz` ✅
- Node: `apache-pulsar-client-node-{version}.tar.gz` ✅
- Python: `pulsar-client-python-{version}.tar.gz` ✅

**Expand row works correctly ✅**
Each row expands to show the full archive directory URL. CPP rows include the "Pre-built DEB, RPM and APK packages..." hint. Go rows include the napi prebuilts hint.

**Node excluded versions ✅**
Older Node versions without a per-version archive directory (`1.0.0`, `1.1.0`, `1.2.0`, `1.3.0`, `1.3.1`, `1.4.0`, `1.3.2`, `1.6.2`, `1.5.0`, `1.4.1`) are correctly filtered out, with a comment in the source explaining why.

**TypeScript ✅**
No new TS errors introduced in `ClientReleasesTable.tsx` or `downloads.tsx`. Pre-existing errors in blog theme files are unrelated to this PR.

**Visual consistency ✅**
Three-column layout (`Release | Source | Release notes`) matches the Older releases table. Zebra striping and row height are consistent.

LGTM from my side. The Node edge-case concern I raised in my previous review has been addressed with the `excludedVersions` filter and the inline comment.

singhvishalkr:
@lhotari The table now follows the expandable row pattern you suggested for CPP and related clients: primary cell links to the source tarball, expanded row exposes the archive directory URL and hints for pre-built artifacts where relevant. @Annaxs re-tested locally on `1820f461` (including the Node `excludedVersions` guard). If this matches what you wanted for the multi-binary case, a fresh review when you have time would unblock the merge queue.

singhvishalkr:
I checked the current head and the client releases table now expands rows for the archive links instead of forcing everything into the main table. That keeps the layout usable and addresses the review concern.

singhvishalkr:
Current head keeps the release links on the tarballs and uses the expandable client-release rows for the directory/prebuilt details. I re-checked the current shape against the earlier review notes, and the page now preserves the file links while still exposing the archive directories when expanded.

singhvishalkr:
Rebased the branch on current `main` in `6349aac51`.

The only conflict was in `src/components/downloads.tsx`; I kept the reviewed client-release table import and upstream's milestone-release helpers. Validation on the rebased head:
- `corepack yarn install --immutable`
- focused TypeScript parse for `src/components/downloads.tsx` and `src/components/ClientReleasesTable.tsx`
- `corepack yarn build`

`corepack yarn typecheck` still reports the existing site-wide `matomoClientModule.ts` and `src/theme/Blog*` type errors; the touched files parse cleanly.

PR #1124, opened 2026-04-22
docs(getting-started-docker-compose): clarify macOS/Windows data-dir step

singhvishalkr:
﻿## Motivation

Fixes #23137.

The docker-compose quickstart currently tells users to run:

```bash
sudo mkdir -p ./data/zookeeper ./data/bookkeeper
# this step might not be necessary on other than Linux platforms
sudo chown -R 10000 data
```

On macOS/Windows (Docker Desktop) the `chown -R 10000` step is both unnecessary -- Docker Desktop's VM handles uid remapping for bind mounts -- **and** actively breaks the stack: `docker compose up` fails with permission errors on `./data/...` because the mounted directory ends up in a state the in-container user can't write.

The existing one-line comment (*"this step might not be necessary on other than Linux platforms"*) is too easy to miss and doesn't tell the user what to do when it already broke.

## Modifications

- Rewrite the Linux instruction so the Linux path is unchanged.
- Add a `:::note macOS and Windows (Docker Desktop)` admonition that:
  - explains *why* chown isn't needed on Desktop hosts,
  - warns that running `sudo chown -R 10000` will typically break `docker compose up`,
  - gives the recovery recipe (delete `./data`, recreate without `chown`, retry).
- Applied to the unversioned `docs/` and the 3 most recent LTS versioned docs (`3.3.x`, `4.1.x`, `4.2.x`). Older versioned docs (2.x, 3.0-3.2, 4.0.x) are intentionally not touched.

## Verifying this change

Render `docs/getting-started-docker-compose.md` &mdash; the Linux snippet must be unchanged; a new admonition should appear after the Linux block explaining the Desktop scenario.

## Does this pull request potentially affect one of the following parts:

- [x] `The docs`

## Documentation

- [x] `doc`

### DIFF
--- a/docs/getting-started-docker-compose.md
+++ b/docs/getting-started-docker-compose.md
@@ -126,13 +126,26 @@ services:
 
 ## Step 2: Create a Pulsar cluster
 
-As preparation, create data directories and change the data directory ownership to uid(10000) which is the default user id used in the Pulsar Docker container.
+As preparation, create the data directories that the `compose.yml` file will bind-mount into the Pulsar containers.
+
+On Linux, the mounted directories need to be owned by uid `10000` -- the default user inside the Pulsar Docker container -- so the containers can write to them:
 
 ```bash
 sudo mkdir -p ./data/zookeeper ./data/bookkeeper
-# this step might not be necessary on other than Linux platforms
 sudo chown -R 10000 data
 ```
+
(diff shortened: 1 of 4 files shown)


lhotari (approved):
LGTM, thanks for improving the docs @singhvishalkr

PR #1126, opened 2026-04-22
docs(administration-stats): clarify messagesConsumedCounter tracks entries, not messages

### LINKED ISSUE #23702: [Doc] The description of messagesConsumedCounter is wrong
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

See https://pulsar.apache.org/docs/next/administration-stats/#topic-internal-stats

messagesConsumedCounter | The number of messages this cursor has acknowledged since this broker loads this topic.
-- | --

Actually the description is wrong. It's the number of **entries**, not messages.

### What is your suggestion?

Fix the description

### Any reference?

_No response_

- [X] I'm willing to submit a PR!

singhvishalkr:
### Summary

`messagesConsumedCounter` in the topic internal stats is documented as "the number of **messages** this cursor has acknowledged since this broker loads this topic", but that's not what the counter tracks.

Looking at the source, the counter is driven by `getNumberOfEntries(...)`:

- `managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedCursorImpl.java` -- `messagesConsumedCounter = -getNumberOfEntries(Range.openClosed(position, ledger.getLastPosition()));`
- `managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/NonDurableCursorImpl.java` -- `messagesConsumedCounter = lastEntryAndCounter.getRight() - initialBacklog;`
- `managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ReadOnlyCursorImpl.java` -- `messagesConsumedCounter = -getNumberOfEntries(Range.closed(readPosition, ledger.getLastPosition()));`

So it counts **managed-ledger entries**. Because an entry can carry a batch of user-visible messages, the counter differs from message count by whatever batching factor the producer used. This matches the inline comment in `NonDurableSubscriptionTest`:
> Note: "topic_internal_stat.cursor.messagesConsumedCounter" means how many messages were acked on this...

(the test file is a bit loose with the wording too, but the assertions all compare against entry counts).

### Change

Rewords the one row in `administration-stats.md` and its three versioned copies:

Before:
> The number of messages this cursor has acknowledged since this broker loads this topic.

After:
> The number of entries (not messages -- an entry can contain a batch of messages) this cursor has acknowledged since this broker loaded this topic.

Closes apache/pulsar#23702.


lhotari (approved):
Thanks!

PR #1127, opened 2026-04-22
docs(node-client): document missing Client configs and defaults

### LINKED ISSUE #24362: [Doc] Client configuration docs changes
### Search before reporting

- [x] I searched in the [issues](https://github.com/apache/pulsar/issues) and found nothing similar.

The docs page for the [Node client](https://pulsar.apache.org/docs/4.0.x/client-libraries-node-configs/#client-configuration) does not seem to document:
 - `useTLS`: boolean
 - `logLevel`: 0, 1, 2, 3 (0 correlates to debug, 1=info, 2=warn, 3=error)

These are, however, captured in the TS types: https://github.com/apache/pulsar-client-node/blob/master/index.d.ts#L21

Further, these defaults are not captured anywhere as far as I can tell:
- Default `tlsTrustCertsFilePath` is `${__dirname}/cert.pem` per the default applied in the JS Client constructor: https://github.com/apache/pulsar-client-node/blob/master/src/Client.js#L24
- Default log level is `info` per the C header file: https://github.com/apache/pulsar-client-node/blob/master/src/Client.h#L56

### What is your suggestion?

I think documenting these things is enough. If there are other places where configurations are inferred or defaulted but not documented, it would be nice to find those and include them, too, but these are the ones I've come across so far.

### Any reference?

_No response_

- [ ] I'm willing to submit a PR!

singhvishalkr:
## Summary

Closes apache/pulsar#24362.

The Node.js client config table in `client-libraries-node-configs.md` was missing several fields that are exposed by `ClientConfig` in [`apache/pulsar-client-node/index.d.ts`](https://github.com/apache/pulsar-client-node/blob/master/index.d.ts). This PR adds rows for:

- `tlsCertificateFilePath` / `tlsPrivateKeyFilePath` -- client-side mTLS material.
- `listenerName` -- picks an advertised listener when the broker exposes more than one.
- `logLevel` -- defaults to `LogLevel.INFO` per [`src/Client.h`](https://github.com/apache/pulsar-client-node/blob/master/src/Client.h) (`pulsar_logger_level_t::pulsar_INFO`).
- `connectionTimeoutMs`.

It also surfaces the previously-undocumented default for `tlsTrustCertsFilePath`: when left unset, the JS wrapper in [`src/Client.js`](https://github.com/apache/pulsar-client-node/blob/master/src/Client.js) falls back to a bundled `cert.pem` seeded from Node.js's `tls.rootCertificates` at install time via `Client.genCertFile()`. Callers who hit certificate errors in restricted environments often rediscover this the hard way; documenting it should save a few debugging cycles.

### Note on `useTLS`

The filing issue mentioned `useTLS` as undocumented, but `useTLS` is not present in the current `ClientConfig` interface on master, so I've not added a row for it. Happy to add one if it's still accepted by the native binding under a different name.

## Scope

- `docs/client-libraries-node-configs.md` (tip / unreleased).
- `versioned_docs/version-4.2.x/client-libraries-node-configs.md` (latest stable).

Older versioned copies are left untouched -- some of the newly-documented fields (`connectionTimeoutMs`, `listenerName`) may not have existed in every older release; happy to backport on request once a maintainer confirms the earliest version.

## Test Plan

- Table columns rendered manually against docusaurus `yarn start` locally.
- Field names + defaults cross-checked against [`index.d.ts`](https://github.com/apache/pulsar-client-node/blob/master/index.d.ts) and [`src/Client.h`](https://github.com/apache/pulsar-client-node/blob/master/src/Client.h) on `pulsar-client-node@master`.

### DIFF
--- a/client-libraries/node-configs.md
+++ b/client-libraries/node-configs.md
@@ -18,11 +18,16 @@ The following configurable parameters are available for Pulsar Node.js clients:
 | `ioThreads` | The number of threads to use for handling connections to Pulsar [brokers](pathname:///docs/reference-terminology#broker). | 1 |
 | `messageListenerThreads` | The number of threads used by message listeners ([consumers](#consumers) and [readers](#readers)). | 1 |
 | `concurrentLookupRequest` | The number of concurrent lookup requests that can be sent on each broker connection. Setting a maximum helps to keep from overloading brokers. You should set values over the default of 50000 only if the client needs to produce and/or subscribe to thousands of Pulsar topics. | 50000 |
-| `tlsTrustCertsFilePath` | The file path for the trusted TLS certificate. | |
+| `tlsTrustCertsFilePath` | The file path for the trusted TLS certificate. If unset, the client falls back to a `cert.pem` file bundled with the package that is seeded from the Node.js runtime's root CAs at install time. | Bundled `cert.pem` |
+| `tlsCertificateFilePath` | The file path for the client TLS certificate used for mTLS. | |
+| `tlsPrivateKeyFilePath` | The file path for the client TLS private key used for mTLS. | |
 | `tlsValidateHostname` | The boolean value of setup whether to enable TLS hostname verification. | `false` |
 | `tlsAllowInsecureConnection` | The boolean value of setup whether the Pulsar client accepts untrusted TLS certificate from broker. | `false` |
 | `statsIntervalInSeconds` | Interval between each stat info. Stats is activated with positive statsInterval. The value should be set to 1 second at least | 600 |
-| `log` | A function that is used for logging. | `console.log` |
+| `listenerName` | The listener name the client will use when connecting to the broker. Useful when [advertising multiple endpoints](pathname:///docs/concepts-multi-tenancy). | |
+| `log` | A function that is used for logging. Receives `(level, file, line, message)` arguments. | `console.log` |
+| `logLevel` | The log level for client-emitted logs. Accepts `LogLevel.DEBUG` (0), `LogLevel.INFO` (1), `LogLevel.WARN` (2), or `LogLevel.ERROR` (3). | `LogLevel.INFO` |
+| `connectionTimeoutMs` | Duration (in milliseconds) to wait for a broker connection to establish before failing. | |
 
 
 ## Producer configs


lhotari (approved):

lhotari:
> Older versioned copies are left untouched -- some of the newly-documented fields (`connectionTimeoutMs`, `listenerName`) may not have existed in every older release; happy to backport on request once a maintainer confirms the earliest version.

Please update 4.0.x docs if possible.

singhvishalkr:
Backported to 4.0.x in `326a189a`, but with one deliberate omission worth flagging before merge.

Pulsar 4.0.x overlapped with `pulsar-client-node` v1.13.0 - v1.13.2 (roughly Aug 2024 - early 2025). I checked `index.d.ts` on each of those tags against the tip interface, and the fields split like this:

- Present in v1.13.x and backported: `tlsTrustCertsFilePath` default, `listenerName`, `logLevel`, updated `log()` callback signature.
- Added in v1.13.2 and backported with a version note on the row: `connectionTimeoutMs`.
- **Not present in the v1.13.x `ClientConfig` interface, so deliberately omitted from 4.0.x**: `tlsCertificateFilePath`, `tlsPrivateKeyFilePath`. They only show up in later client releases, so documenting them on 4.0.x would describe fields that the contemporaneous binding wouldn't accept. Happy to switch the call if you'd rather keep the row for forward-compat (users who upgrade the client under a 4.0.x server).

Refs for the version walk:
- https://raw.githubusercontent.com/apache/pulsar-client-node/v1.13.0/index.d.ts
- https://raw.githubusercontent.com/apache/pulsar-client-node/v1.13.2/index.d.ts
- https://raw.githubusercontent.com/apache/pulsar-client-node/v1.14.0/index.d.ts

lhotari:
> Pulsar 4.0.x overlapped with `pulsar-client-node` v1.13.0 - v1.13.2 (roughly Aug 2024 - early 2025). I checked `index.d.ts` on each of those tags against the tip interface, and the fields split like this:

pulsar-site currently mistakenly contains client documentation in versioned docs for Pulsar. Since newer clients are compatible with older broker versions for most features, the client documentation should be maintained only for the most recent release. This is something that we'd need to address separately.
The `client-libraries` should be moved to top level and https://github.com/apache/pulsar-site/blob/main/static/.htaccess should be used to redirect from old urls to the top level client-libraries location.

singhvishalkr:
Thanks for flagging the bigger picture. Happy to help on the client-libraries lift to top-level + `.htaccess` redirects whenever you're ready to tackle it -- just ping me on the tracking issue and I can take a first pass at the move/redirect mapping.

lhotari:
I created #1131. I already have the plan Claude Code, so I'll put it forward.

lhotari:
I've created #1132. Once that is merged, please apply the changes to the new location.

singhvishalkr:
Ack, watching #1132. Once it's merged I'll rebase this onto `main`, move the Node-client additions into the new `docs/client-libraries/` location, and drop the versioned-docs copies from here so the diff matches the new structure.

lhotari:
@singhvishalkr The PR to relocate client library docs has been merged.

singhvishalkr:
Rebased onto main after #1132 landed. Dropped the `docs/client-libraries-node-configs.md` and the two `versioned_docs/` copies; the six additions now sit on the unversioned `client-libraries/node-configs.md` instead:

- `tlsTrustCertsFilePath` default: noted the bundled `cert.pem` fallback.
- New rows for `tlsCertificateFilePath`, `tlsPrivateKeyFilePath`, `listenerName`, `logLevel`, `connectionTimeoutMs`.
- `log` description expanded with the `(level, file, line, message)` signature.

Since the new location is unversioned, the 4.0.x-specific note on `connectionTimeoutMs` availability no longer applies here — if a per-version caveat becomes relevant again the `feature-matrix.mdx` is probably the better place for it. One commit, 1 file, +7/-2.

lhotari (approved):
LGTM, thanks for the contribution!

PR #1176, opened 2026-06-18
Document the V5 Java client SDK alongside the current SDK

merlimat:
### Motivation

Pulsar 5.0 adds the V5 Java client (`pulsar-client-v5`) for [scalable topics](https://github.com/apache/pulsar/blob/master/pip/pip-466.md), but the Java client docs only cover the current SDK (`pulsar-client`). This adds V5 client documentation and clarifies the relationship between the two SDKs, while keeping the existing (current) SDK docs and examples unchanged.

### Modifications

- **New `client-libraries/java-v5.md`** — V5 client guide: install (`pulsar-client-v5`, Java 17), create a client, produce, the three consumer types (Stream / Queue / Checkpoint), schemas, and transactions.
- **`client-libraries/java.md`** — added a "Java client SDKs" section with a side-by-side comparison of the current vs V5 SDK (package, artifact, topic types, consumption model, minimum Java).
- **`sidebarsClientLibraries.js`** — added the V5 page under the Java client category.

The existing current-SDK pages (set up / initialize / use / tracing) are unchanged. The V5 code examples were verified against the actual `pulsar-client-api-v5` interfaces.

### Notes

Independent of the scalable-topics and messaging-rework PRs — the cross-links to those concept pages use `pathname:///docs/...`, so there is no build dependency and this can merge in any order.

PR #1177, opened 2026-06-18
Add a guide for migrating Java applications to the V5 client

merlimat:
### Motivation

With the V5 Java client now documented (#1176), application developers need guidance for moving existing code from the current client API to the V5 API. This adds that migration guide.

### Modifications

- **New `client-libraries/java-migrate-to-v5.md`** ("Migrate to V5"):
  - When (and when not) to migrate; the two SDKs coexist in one JVM, so migration is incremental.
  - An **API mapping table** — the subscription-type collapse (Exclusive/Failover → Stream, Shared/Key_Shared → Queue, `Reader` → Checkpoint), `Schema.STRING` → `Schema.string()`, acknowledgment changes, `Duration`/`Instant`, and config records.
  - **Side-by-side current-vs-V5 code** for the client, producer, and all three consumer migrations, plus schemas/transactions/configuration and the server-side `pulsar-admin scalable-topics migrate` step.
- **`sidebarsClientLibraries.js`** — adds the page under Java client, after the V5 client guide.

The V5 examples were verified against the `pulsar-client-api-v5` interfaces (e.g. the `Reader` → Checkpoint section notes that a saved `MessageId` is not interchangeable with a `Checkpoint`).

### Notes

Independent PR off `main`, building on the merged #1176.

### DIFF
--- a/client-libraries/java-migrate-to-v5.md
+++ b/client-libraries/java-migrate-to-v5.md
@@ -17,14 +17,72 @@ You don't have to migrate. The current SDK is fully supported and remains the ri
 
 The two SDKs are independent and can run **side by side in the same JVM**, so you can migrate incrementally -- one producer or consumer at a time -- rather than all at once. A typical path:
 
-1. Add the `pulsar-client-v5` dependency alongside your existing `pulsar-client`.
+1. Add the `pulsar-client-v5` dependency (plus `pulsar-client-original` for the not-yet-migrated v4 code -- see [Dependencies](#dependencies)).
 2. Move producers and consumers to the V5 API. The V5 client works against your **existing** `persistent://` topics, so you can do this without changing any topic.
 3. When you're ready, [migrate the topics themselves to scalable topics](#migrating-the-topics) -- a separate, server-side step that is transparent to V5 applications.
 
 ## Prerequisites
 
 - **Java 17.** The V5 client requires Java 17 (the current SDK supports Java 8+).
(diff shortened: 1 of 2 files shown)


client-libraries/java-migrate-to-v5.md
```diff
@@ -0,0 +1,210 @@
...
+
+You don't have to migrate. The current SDK is fully supported and remains the right choice for non-Java applications, for applications that don't need scalable topics, and for non-persistent topics (which the V5 client does not support). Migrate a Java application when you want scalable topics or the V5 API.
+
+:::
+
+## How migration works
+
+The two SDKs are independent and can run **side by side in the same JVM**, so you can migrate incrementally -- one producer or consumer at a time -- rather than all at once. A typical path:
+
+1. Add the `pulsar-client-v5` dependency alongside your existing `pulsar-client`.
```
lhotari: Since `pulsar-client-v5` depends on `pulsar-client-original`, the `pulsar-client` dependency should be removed completely or replaced with `pulsar-client-original`. 

In addition, it is strongly recommended to use pulsar-bom for aligning Pulsar client library dependencies and netty-bom for aligning Netty library dependencies. (an example for both Maven and Gradle builds would be very useful in clarifying this. There's currently examples of pulsar-bom at https://pulsar.apache.org/docs/client-libraries/java-setup/#pulsar-bom.

lhotari (commented):
Added a review comment

